Issue #134, opened 2014-01-07
clear_output?


lucasb-eyer:
I've been successfully using `IPython.display.clear_output` for some animations[1] in python. Now, I'm trying to replicate this in IJulia. I tried both

```julia
@pyimport IPython.display as dy

display("text/html", "<b>Bye :(</b>")
dy.clear_output(true)
display("text/html", "<b>Hello!</b>")
```

and

```julia
display("text/html", "<b>Bye :(</b>")
send_ipython(publish, msg_pub(execute_msg, "clear_output", [:wait => true]))
display("text/html", "<b>Hello!</b>")
```

but none seem to work, i.e. I always see both "**Bye :(**" and "**Hello!**" as output.

The second idea comes from mirroring what the `display` methods do and looking at what IPython itself does, specifically `clear_output` in [zmqshell.py:91](https://github.com/ipython/ipython/blob/master/IPython/kernel/zmq/zmqshell.py#L91)

When turning on debugging in IJulia, I can see the attached output in the commandline, which looks "good" to me. (I have, also tried with `wait` set to `false` and `"wait"` instead of `:wait`, all the same.)

Any ideas about what's going on?

```
SENDING IPython Msg [ idents pyin ] {
  header = ["msg_type"=>"pyin","msg_id"=>"c8427e60-58de-4bb6-a131-18d72396a2d5","username"=>"username","session"=>"89FE998B22494671A572701C760E0D78"],
  metadata = Dict{String,Any}(),
  content = ["execution_count"=>124,"code"=>"import IJulia: set_verbose, send_ipython, publish, msg_pub, execute_msg, display_dict, displayqueue, undisplay\nset_verbose()\n\ndisplay(\"text/html\", \"<b>Bye :(</b>\")\nsend_ipython(publish,\n             msg_pub(execute_msg, \"display_data\",\n                     [\"source\" => \"julia\", # optional\n                      \"metadata\" => Dict(), # optional\n                      \"data\" => [\"text/html\" => stringmime(MIME(\"text/html\"), \"<em>trololol?</em>\")] ]))\nsend_ipython(publish, msg_pub(execute_msg, \"clear_output\", [:wait => true]))\n#dy.clear_output(true)\ndisplay(\"text/html\", \"<b>Hello!</b>\")"]
}
SENDING IPython Msg [ idents status ] {
[...]

PR #165, opened 2014-04-21
clear_output support (not working yet) for #134

JustAJoe:


stevengj:
Just pushed a working (in IPython 3) and rebased version of this PR.  e.g.
```jl
for i = 1:1000
    IJulia.clear_output(true)
    display(i)
end
```
works and animates the display.

cedeerwe:
I cannot get this working in JuliaBox, neither with the 0.3.9 kernel nor 0.4.0 dev kernel with the error being clear_output not defined.

stevengj:
That's because it requires the IJulia `master` version, and probably JuliaBox is using the last tagged release.

PR #210, opened 2014-07-26
add kernelspec for IPython 3.0

Carreau:
~~Still WIP~~, 

~~But once https://github.com/ipython/ipython/pull/6126 is merge this will allow to change running kernel in a single notebook server.~~ This add the minimal changes necessary to register a new kernel with IPython.

~~~The things that are still to figure out are the modification Julia bring to tooltip, and the logo/favicon that does not change (yet).~~

![capture decran 2014-07-26 a 22 43 23](https://cloud.githubusercontent.com/assets/335567/3712236/8a881e10-1505-11e4-8ecb-88739b503de0.png)

[edit] kernel selector is now in the `Kernel > change kernel` Menu.


rudimk:
Following this thread. Would probably then use this to get the Torch kernel to run with Jupyter.

deps/build.jl
```diff
@@ -130,6 +143,36 @@ copy_config("custom.js", joinpath(juliaprof, "static", "custom"))
...
+#######################################################################
+
+
+if ipyvers >= v"3.0-"
+    eprintln("Found IPython version $ipyvers ... installing kernelspec.")
+
+    juliakspec = joinpath(chomp(readall(`$ipython locate`)),"kernels","julia")
+    ks = @compat Dict(
+        "argv" => kernelcmd_array,
+        "display_name" => "Julia "*string(VERSION),
```
minrk: should add `"language" => "julia"`

Carreau:
Updated and rebased on master. In conjunction with #276 should make IJulia work with IPython 3.0, either wether or not you use `--profile=julia`.

capture d ecran 2015-02-17 a 15 17 53
capture d ecran 2015-02-17 a 15 17 12

StefanKarpinski:
I'm super excited about this feature.

Carreau:
> I'm super excited about this feature.

Yeah, Someone a little more experienced than me should probably have a look at implementing message spec v5. That would allow to do some nifty things with completion like:

`atype)<tab>` -> actually complete in front to `function(atype)` which would give the "seamless" tab introspection on doted accessors like in other languages, or whatever your kernel can do.

stevengj:
I'll merge it if you think it's ready to go.

deps/build.jl
```diff
@@ -125,11 +138,55 @@ copy_config("ijuliafavicon.ico",
 # custom.js can contain custom js login that will be loaded
 # with the notebook to add info and/or monkey-patch some javascript
 # -- e.g. we use it to add .ipynb metadata that this is a Julia notebook
-copy_config("custom.js", joinpath(juliaprof, "static", "custom"))
+if ipyvers <= v"3.0-"
```
stevengj: Why is there an `if` statement here if both branches of the `if` do the same thing?

deps/build.jl
```diff
@@ -125,11 +138,55 @@ copy_config("ijuliafavicon.ico",
...
+  ## shiped with IJulia ?
+  copy_config("custom.js", joinpath(juliaprof, "static", "custom"))
+end
 
 # julia.js implements a CodeMirror mode for Julia syntax highlighting in the notebook.
 # Eventually this will ship with CodeMirror and hence IPython, but for now we manually bundle it.
 
-copy_config("julia.js", joinpath(juliaprof, "static", "components", "codemirror", "mode", "julia"))
+if ipyvers <= v"3.0-"
+    copy_config("julia.js", joinpath(juliaprof, "static", "components", "codemirror", "mode", "julia"))
```
stevengj: Can't we do something like what you did with `custom.js` to make `julia.js` work with IPython 3?

Suppose someone has IPython 2.0 installed and installs IJulia (hence gets `julia.js`) then upgrades to IPython 3.  Won't the `julia.js` file lying around cause things to break?

deps/build.jl
```diff
@@ -125,11 +138,55 @@ copy_config("ijuliafavicon.ico",
...
+end
 
 #######################################################################
+#       Part specific to Jupyter/IPython 3.0 and above
+#######################################################################
+
+
+if ipyvers >= v"3.0-"
```
stevengj: Can't we just install the kernelspec unconditionally?  After all, the user may subsequently upgrade to IPython 3, and these files should be harmless under IPython 2, no?

rgbkrk: That's a good point.

deps/build.jl
Carreau: cause it's todo to remove on 3.0, but I can remove the `if`.

deps/build.jl
Carreau: If you copy `julia.js` it will take precedence. I'm fine with it, but codemirror julia.js seem to be more recent than IJulia one by a few month (march, vs october). So on 3.0 I would be enclined to not-copy-it, or even remove it if exist. 
The question is how longer do you plan to have a copy of Julia.js in IJulia.

deps/build.jl
Carreau: Sure, done, I prefer to keep things clean when unneeded.

stevengj:
Thanks so much for your efforts on this, @Carreau.  Now that it's merged, the people on the bleeding-edge IJulia can bang on it a little bit, and then I can tag a new release.

Carreau:
Sure, I hope to find some time to do a bit more of Julia myself. Might demo
that at Strata tomorrow. Thanks !

On Thu, Feb 19, 2015 at 7:02 PM, Steven G. Johnson <notifications@github.com

PR #211, opened 2014-08-06
WIP: Integrate IPython setup on windows

davidanthoff:
I tested this on a brand new windows system with nothing but julia installed:

````julia
Pkg.clone("https://github.com/davidanthoff/IJulia.jl.git","IJulia")
Pkg.checkout("IJulia", "integratedipysetup")
Pkg.build("IJulia")

using IJulia
notebook()
````

works on that system.

This PR lacks proper error handling etc, consider it more a proof of concept.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -9,6 +9,58 @@ eprintln(x...) = println(STDERR, x...)
 
 juliaprofiles = Array(String,0)
 
+@windows_only begin
+    ipython_version_to_install = "2.4.1"
+
+    existing_install_tag_filename = normpath(pwd(),"usr","python342-exists")
+    downloadsdir = normpath(pwd(), "downloads")
+    pythonzipfilename = normpath(pwd(), "downloads", "python-3.4.2.zip")
+    pyinstalldir = normpath(pwd(),"usr","python34")
+    pythonexepath = normpath(pwd(),"usr","python34","python.exe")
+    ijuliaprofiledir = normpath(pwd(), "usr", ".ijulia")
+
+    upgrade_private_python = ispath(existing_install_tag_filename)
+
+    if upgrade_private_python
+        rm(existing_install_tag_filename)
+
+        run(`$pythonexepath -m pip install -U pip`)
+        run(`$pythonexepath -m pip install -U ipython[notebook]==$ipython_version_to_install`)
+    else
+        using BinDeps
+
+        if ispath(downloadsdir)
+            rm(downloadsdir, recursive=true)
+        end
+
+        if ispath(normpath(pwd(),"usr"))
+            rm(normpath(pwd(),"usr"), recursive=true)
+        end
+
+        mkdir(downloadsdir)
+
+        run(download_cmd("https://sourceforge.net/projects/minimalportablepython/files/python-3.4.2.zip", "$pythonzipfilename"))
+
+        run(`7z x $pythonzipfilename -y -o$pyinstalldir`)
+
+        run(`$pythonexepath -m ensurepip`)
+        run(`$pythonexepath -m pip install -U pip`)
+
+        run(`$pythonexepath -m pip install ipython[notebook]==$ipython_version_to_install`)
+    end
+
+    if ispath(ijuliaprofiledir)
+        rm(ijuliaprofiledir, recursive=true)
+    end
+
+    run(`$pythonexepath -m IPython profile create --ipython-dir="$ijuliaprofiledir"`)
+
+    internaljuliaprof = chomp(readall(`$pythonexepath -m IPython locate profile --ipython-dir="$ijuliaprofiledir"`))
+    push!(juliaprofiles, internaljuliaprof)
+
+    touch(existing_install_tag_filename)
+end
+
 include("ipython.jl")
 const ipython, ipyvers = find_ipython()
 
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -7,28 +7,38 @@ using Compat
 # print to stderr, since that is where Pkg prints its messages
 eprintln(x...) = println(STDERR, x...)
 
+juliaprofiles = Array(String,0)
+
 include("ipython.jl")
 const ipython, ipyvers = find_ipython()
 
-if ipyvers < v"1.0.0-dev"
-    error("IPython 1.0 or later is required for IJulia, got $ipyvers instead")
+if ipython==nothing
+    if length(juliaprofiles)==0
+        error("IPython 1.0 or later is required for IJulia")
+    else
+        eprintln("No system IPython found, using private IPython.")
+    end
+elseif ipyvers < v"1.0.0-dev"
+    if length(juliaprofiles)==0
+        error("IPython 1.0 or later is required for IJulia")
+    else
+        eprintln("IPython 1.0 or later is required for system IJulia, got $ipyvers instead. Skipped integration with system IPython.")
+    end
 else
     eprintln("Found IPython version $ipyvers ... ok.")
+
+    # create julia profile (no-op if we already have one)
+    eprintln("Creating julia profile in IPython...")
+    run(`$ipython profile create julia`)
+
+    systemjuliaprof = chomp(readall(`$ipython locate profile julia`))
+    push!(juliaprofiles, systemjuliaprof)
 end
-
-#######################################################################
-# Create Julia profile for IPython and fix the config options.
-
-# create julia profile (no-op if we already have one)
-eprintln("Creating julia profile in IPython...")
-run(`$ipython profile create julia`)
-
-juliaprof = chomp(readall(`$ipython locate profile julia`))
 
 # set c.$s in prof file to val, or nothing if it is already set
 # unless overwrite is true
-function add_config(prof::String, s::String, val; overwrite::Bool=false)
-    p = joinpath(juliaprof, prof)
+function add_config(profdir::String, prof::String, s::String, val; overwrite::Bool=false)
+    p = joinpath(profdir, prof)
     r = Regex(string("^[ \\t]*c\\.", replace(s, r"\.", "\\."), "\\s*=.*\$"), "m")
     if isfile(p)
         c = readall(p)
@@ -65,43 +75,6 @@ c.$s = $val
     end
 end
 
-# add Julia kernel manager if we don't have one yet
-if VERSION >= v"0.3-"
-    binary_name = @windows? "julia.exe":"julia"
-else
-    binary_name = @windows? "julia.bat":"julia-basic"
-end
-
-kernelcmd_array = [escape_string(joinpath(JULIA_HOME,("$binary_name")))]
-
-if VERSION >= v"0.3"
-    push!(kernelcmd_array,"-i")
-end
-
-append!(kernelcmd_array, ["-F", escape_string(joinpath(Pkg.dir("IJulia"),"src","kernel.jl")), "{connection_file}"])
-
-
-
-kernelcmd = JSON.json(kernelcmd_array)
-
-add_config("ipython_config.py", "KernelManager.kernel_cmd",
-           kernelcmd,
-           overwrite=true)
-
-# make qtconsole require shift-enter to complete input
-add_config("ipython_qtconsole_config.py",
-           "IPythonWidget.execute_on_complete_input", "False")
-
-add_config("ipython_qtconsole_config.py",
-           "FrontendWidget.lexer_class", "'pygments.lexers.JuliaLexer'")
-
-# set Julia notebook to use a different port than IPython's 8888 by default
-add_config("ipython_notebook_config.py", "NotebookApp.port", 8998)
-
-#######################################################################
-# Copying files into the correct paths in the profile lets us override
-# the files of the same name in IPython.
-
 rb(filename::String) = open(readbytes, filename)
 eqb(a::Vector{Uint8}, b::Vector{Uint8}) =
     length(a) == length(b) && all(a .== b)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by davidanthoff** [comment]
This is currently blocked on JuliaLang/Julia#9671. @tkelman, @ihnorton, @vtjnash, @Keno, @StefanKarpinski any help with that bug would be appreciated, fixing that one is beyond my abilities.

**Comment by stevengj** [comment]
Great, glad that problem got fixed.

**Comment by davidanthoff** [comment]
This is now ready for others to try out and look at. The instructions at the top of this PR on how to are still valid and should work.

There is a non-trivial amount of change to the build.jl script, so I'm quite nervous about breaking things. So please, try this out on as many different systems as you can before we merge this and a very, very thorough code review would also be good. I myself tested this on a fresh WinXP system with Julia 0.3.4 (so that is x86) and a fresh Windows 8.1 x64 system, again with Julia 0.3.4. I tested both platforms with and without a system IPython (Anaconda) installed. I did NOT test any non-windows platform.

The general strategy of this is now as follows: builds.jl on windows does the following steps:
- delete any previous downloads, python installs and private profile directories
- downloads miniconda and pip
- installs a fresh python system via the miniconda installer into IJulia/deps/usr/python34
- installs pip into that system
- uses pip to install IPython and its dependencies
- configures this "private" IPython for IJulia
- searches for any "system" IPython that might already be on the system and if it finds one configures that one for IJulia as well

The ``notebook()`` function on windows now always starts the private IPython install, and that is also the only user visible entry point into the private IPython install.

Here is a list of design decisions that I could also see going some other way:
- the system creates a .ijulia directory with the profile under IJulia/deps/usr. I guess that directory could also go into the home directory, parallel to the normal .ipython directory. My sense right now is that I prefer it to be hidden in the normal IJulia directory, so that there is less room for people to mess up this private IPython setup. E.g. if one deletes the IJulia package, everything is gone and there is no other place on the system where one might have to delete things as well.
- The build.jl is inefficient in that it downloads and installs a new IPython with every run. There could be a lot of caching etc that would improve performance. My sense for now is that this is something one could improve later on
- The profile directory for the private IPython install now gets deleted and recreated with every ``Pkg.build``. I like that because it makes the system less brittle, but it does mean that there is no way to do any customization of the private IJulia install. My sense is that for people who want to mess around with settings, they should just install a system IPython and do their customization there.
- The build.jl hard codes specific versions of IPython and its dependencies for the pip install. So this won't automatically get the latest IPython etc. I prefer this because again it gives us more stability. But it does mean that the IJulia maintainers need to periodically update the build.jl script to get the newest IPython versions, after they have been tested.

I think at this point a lot of people should look at this because this would impact a very hot code path (build.jl) for a lot of people: @stevengj, @tkelman, @StefanKarpinski, @vtjnash, @Keno, @ihnorton, @Carreau.

**Comment by stevengj** [comment]
Why do you use `pip` rather than `conda` to install IPython?  Since you have miniconda, can't you just do `conda install ipython`?

**Comment by davidanthoff** [comment]
I still have some vague hope that at some point Python itself will release an installer that we could use instead of the miniconda installer, i.e. that the miniconda installer could be replaced with the normal Python installer... I'm already not very happy to take a dependency on miniconda, but I just couldn't get the normal python setup to work. But the more we stick to the "official" python stack the better it seems to me?

Would you see a benefit of using conda here instead of pip?

**Comment by stevengj** [comment]
If we are using miniconda anyway, I'd be much happier sticking with the Anaconda IPython package.  For one thing, we don't need to specify a version; it will be reasonably up to date.  For another thing, we don't need to mess around with specifying the dependencies.

Can you remind me why you need `CreateProcess` and can't just call `run`?

**Comment by stevengj** [comment]
If there is a global `ipython` already installed, I'd like to keep installing a profile there too, in addition to the private install.

**Comment by davidanthoff** [comment]
> If we are using miniconda anyway, I'd be much happier sticking with the Anaconda IPython package. For one thing, we don't need to specify a version; it will be reasonably up to date. For another thing, we don't need to mess around with specifying the dependencies, and we can be reasonably sure that Anaconda picks dependencies that work together.

My understanding is that the pypi package of IPython is maintained by the IPython org, so if we use pip we are getting things straight from them, one less group (the anaconda people) involved and I am sure things will be as up-to-date as they can get. The pip setup is actually the first one described on the ipython.org website, so I would be very surprised if they didn't make sure it works all the time. The whole dependency thing that anaconda really solves is actually of little concern to us here, the really tricky things really all just seem to be around numpy and friends, I don't think IPython has anything that requires something like anaconda.

Whether we specify a version during the install is not a question conda vs pip, pip also supports a command that installs the latest version. Same for manual dependencies, there is a pip command that install the latest ipython with all the necessary dependencies.

The reason I opted to not put in any "get the latest ipython with deps" (either with conda or pip) is that I think it would introduce a potential instability. The specific scenario I have in mind is that a new IPython version is released that is in some way incompatible with IJula, and then people would get that version automatically and things would not work. I would prefer that someone from the IJulia team tries out a new IPython version at least once with IJulia, checks whether it works, and then controls when that new IPython version is used as the private IJulia bits. With the current code the IJulia devs control EXACTLY which bits get used for the private install, and I think that is a good thing in terms of robustness. Relying on the maintainers of either a pypi or anaconda package for that choice means it is made by someone who is not testing things with IJulia...But in any case, this would not be a reason to use conda, pip can do the same thing.

Another reason I didn't use conda is that I don't understand it well, but I do understand pip very well. We could probably get this to work with conda as well, but I don't see any benefit, only costs in terms of figuring out how that works, more testing etc... I'm lazy, sorry, and testing this on the various configurations already took way too much time :)

> Can you remind me why you need CreateProcess and can't just call run?

The miniconda setup requires that the path where it should be installed is passed on without any quotes, even if there is a space in the path... As far as I understand ``run`` would put quotes around things.

> If there is a global ipython already installed, I'd like to keep installing a profile there too, in addition to the private install.

That is what the current code is doing (or should do).

**Comment by tkelman** [comment]
Semi-aside, are people aware of the current status of Jupyter and whether any design changes will be made that could simplify this for us? I kind of wish this could just be a standard portable C library that all the various language bindings could use...

**Comment by stevengj** [comment]
Jupyter is and will be in Python for the foreseeable future.

**Comment by stevengj** [comment]
@davidanthoff, I don't necessarily want the latest, most bleeding-edge version.  I want something that has been tested to work together, and keeps up with important bugfixes and similar patches.   That's the value added by system integrators like Anaconda.

My preference would be to just use the latest Anaconda version of IPython and see how that goes; I think this will be a lot less headaches for us in terms of keeping track of all the bugfixes and other patch releases of all the dependencies. Usually, it takes a few weeks between an IPython release and the corresponding Anaconda package update, so that gives us time to test things.  Moreover, a lot of people use IJulia with Anaconda (because they want to call other Python packages via PyCall), so we need to try hard to keep that working anway.

As for putting quotes around things, my understanding is that `run` does not use the shell at all and circumvents quoting entirely.  This is, in fact, the whole point of `run`.  See [this blog post by Stefan](http://julialang.org/blog/2012/03/shelling-out-sucks/).   If there is a case where this is not working, it should be fixed.

**Comment by Carreau** [comment]
> Jupyter is and will be in Python for the foreseeable future. We're still going to need it installed somewhere.

Technically Jupyter should be only the spec.
Note that we don't reject other language for the core, I would more consider Jupyter/IPython as a reference implementation (hopefully we won't have the same problem that Python == CPython). 
We are just majority of Python devs. Also we will split the repo to encourage non-python dev to participate to the core. 

For example someone is writing a [java replacement](https://github.com/jove-sh/jove) (AFAIU).

We try to make the javascript/html part as much separated from the Python side than possible, 
so re-writing the notebook python frontend  part should be a doable task. It is "only" 9k lines, comment docstring and blankline included. Removing backend widgets, and test. 

```
~/ipython (master) $ cat IPython/html/**/*.py| grep -v widgets | grep -v tests | wc -l
    9290
```

As we **will** split IPython in many part, the installation of the server should at least be much smaller, 
and might not rely on some historical things that make IPython not so easy to install. 

Getting late here, but will come bak to this thread after some rest tomorrow.

**Comment by davidanthoff** [comment]
> As for putting quotes around things, my understanding is that run does not use the shell at all and circumvents quoting entirely. This is, in fact, the whole point of run. See this blog post by Stefan. If there is a case where this is not working, it should be fixed.

See this discussion from above:

>> The single quotes don't get passed to the operating system, they just denote which arguments are a unit.

> except that this is windows, so they do get passed out to the operating system (although they will turn into double quotes " and other quotes will be properly escaped)

**Comment by stevengj** [comment]
@davidanthoff, as @Keno noted, the problem is that you normally shouldn't use quotes at all in the `run` command.  Quotes are a shell thing, and apparently was also some weirdness with `msiexec` (thank you, Microsoft).  But since now you are no longer using `msiexec` maybe you can just go back to using `run` and omitting all internal quotes.

**Comment by davidanthoff** [comment]
@stevengj, the problem is that
````
x = "Two parts"
run(`cmd $x`)
````
will put quotes around "Two parts" on windows, as it should for well behaved windows programs. But neither the miniconda installer nor the msiexe program are well behaved.

The stuff in wincall.jl should probably go into a WindowsAPI.jl package at some point.

**Comment by stevengj** [comment]
I'm confused, why should it put quotes around the arguments for well-behaved Windows programs?

**Comment by davidanthoff** [comment]
See above:

> libuv quotes the arguments under the assumption that the recipient program will use the correct windows API function to unquote them. perhaps the authors of msiexec are unfamiliar with windows? anyways, there's a flag we should expose in the libuv api that lets you turn this off

libuv here assumes that programs follow [these](http://msdn.microsoft.com/en-us/library/a1y7w461.aspx) guidelines, but both msiexec and the miniconda installer don't.

I don't know if someone has by now exposed that libuv flag that @vtjnash mentions above, I don't understand that part of the library well enough to do it.

**Comment by vtjnash** [comment]
this flag (https://github.com/joyent/libuv/blob/52ae456b0d666f6f4dbb7f52675f4f131855bd22/include/uv.h#L871-L875) would need to be set here (https://github.com/JuliaLang/julia/blob/110108631e42d1a47f741091358b549ebd4b1991/src/jl_uv.c#L335-L367) when requested by the user

**Comment by stevengj** [comment]
I still don't understand this.  Why is libuv passing a command-line that gets parsed at all?  Can't it call something like [_wspawnlp](http://msdn.microsoft.com/en-us/library/t7y6z6fd.aspx) and pass the arguments directly, rather than as a single command line that needs to be parsed somehow?

**Comment by vtjnash** [comment]
internally, libuv/julia is already using CreateProcess, but tries to provide some sanity in passing command line arguments, stdio pipes, and avoiding the whitespace-related security attacks mentioned in the documentation http://msdn.microsoft.com/en-us/library/windows/desktop/ms682425%28v=vs.85%29.aspx

`_wspawnlp` internally calls `CreateProcess` (http://msdn.microsoft.com/en-us/library/20y988d2.aspx), but is wrapping it to provide less functionality

**Comment by davidanthoff** [comment]
@stevengj, I tried the ``conda install ipython`` option. First it gives a warning that it tries to update conda itself, but that the conda process is running. Second, the thing that gets installed doesn't work, it crashes with an error when the ipython notebook is started (``ImportError: No module named 'zmq'``). The ``pip`` version works, so I'm kind of hesitant to put more time into switching to ``conda`` at this point.

I'm also not sold on the better system integration tests we would get from using conda. What we really need tested is the combination of 1) the current IJulia, 2) a given IPython on Python 3.4 x86 [this being the private IPython thing my PR is setting up] with any 3) global/system Python setup that users might want to call via PyCall. 3) will almost always be Python 2.7, often from Anaconda, but potentially also Canopy or something else. I very, very much doubt that the anaconda guys are testing how parallel installations of different python versions work together...

Also, apart from the error just described, there are further questions: is conda set up for multiple, parallel installations on the same system? I.e. a private conda setup in the IJulia folder, plus a "public" conda setup from a user anaconda installation. I have no clue, but I do know that that kind of setup is supported by pip. If someone has more experience with conda and knows the answers to those questions we could look further into it, but I'm not familiar enough with ``conda`` to make that call.

If we don't want to manage dependencies of IPython manually in the build.jl script, we can just use this command, it will automatically install all the necessary dependencies:

````
pip install ipython[notebook]==2.3.1
````

if we want the script to just pull the latest IPython always, we can use this:

````
pip install ipython[notebook]
````

Both work, so this is just a question of how much control we want to encode over version in the build.jl script.

I've pushed a commit for now that goes with the first of these two options.

**Comment by Carreau** [comment]
> The profile directory for the private IPython install now gets deleted and recreated with every Pkg.build. I like that because it makes the system less brittle, but it does mean that there is no way to do any customization of the private IJulia install. My sense is that for people who want to mess around with settings, they should just install a system IPython and do their customization there.

We should support multiple profile directory on master (IIRC) like system level and user lever. 
If you like you might be able to get default config re-written, and user config in another place. 

> The build.jl hard codes specific versions of IPython and its dependencies for the pip install. So this won't automatically get the latest IPython etc. I prefer this because again it gives us more stability. But it does mean that the IJulia maintainers need to periodically update the build.jl script to get the newest IPython versions, after they have been tested.

Hardcode only major version ? like 3.x.y or 3.1.t. That would at least relieve of most maintaining.

> [Conda stuff]

I'm not a user of conda myself,  so do as you prefer. Though, conda goal is basically to make isolated environment, and using conda will mean you want to sandbox a mechanisme made to create sandbox which is a bit ironic.

That beeing said, pip and PyPI have been conceived originally to be consumed by other packages manager, so I don't see any reason to treat pip-install as "official"; It is  one that exist on most platfoms, and since python `2.7.9` and `3.3`, is sure to exist when python is installed. Conda need an extra-install step.

I would annoy Continuum IO to know how much they want Julia in anaconda to get some sens of how much they would test Newer version of IPython with newer version of Julia.

**Comment by davidanthoff** [comment]
@Carreau, am I right that you guys are maintaining the wheels for IPython that are up on pypi?

**Comment by davidanthoff** [comment]
New problem: the miniconda silent installation very reliably crashes my Outlook 2013 every time I run it... This has nothing to do with Julia, it also happens when I start the miniconda setup from the command line. It seems that this is a problem with NSIS that others have [had](http://sourceforge.net/p/nsis/bugs/1049/) [before] (http://forums.winamp.com/showthread.php?t=260624). I've opened an issue with the conda folks conda/conda#1084.

Question is what do we do in the meantime... Options:

1. Ignore it, our install works... Maybe put something in the README saying "quit oulook before you do a setup". I really don't like this, we might cause people to lose draft emails and things like that...
2. Wait for the conda people to get the NSIS people to fix this. Might take a long time...
3. Try to go back to my attempts to use the official python.org msi file to do the python setup. This includes the nasty MSI patching. Last time I gave up because of a crash I saw when I called the MSI API, but I think that was probably caused because I didn't know about the stdcall option for ccall, so this might be worth revisiting.
4. The python.org crowd is creating a brand [new installer](http://stevedower.id.au/blog/the-python-3-5-installer/) for Python 3.5. I'm trying to engage them to to add a silent/portable install option to that official installer. If they do that, it would mean that we could use the official python.org installer without any nasty MSI patching. I think the current release date for Python 3.5 is September...

Thoughts?

**Comment by stevengj** [comment]
Option 1 isn't really acceptable.  Might be worth revisiting 3.

**Comment by tkelman** [comment]
I think we're also using NSIS for the base Julia installer, I hope we aren't hitting the same problem there?

**Comment by davidanthoff** [comment]
@tkelman, when I experimented with the chocolatey package I used the silent install option for Julia, and I don't remember any problem.

**Comment by Carreau** [comment]
> @Carreau, am I right that you guys are maintaining the wheels for IPython that are up on pypi?

Yes.

**Comment by davidanthoff** [comment]
I had a short email thread on the python-dev list about a portable python install or the option for a silent portable python option in the default python setup file. Long story short, it is not going to happen anytime soon. The suggestion I got was to just install python on a system, put all the binary files that are installed that way into a zip and distributed these bits with our stuff manually. So that is what the current branch does: it downloads a zip file that has a complete python, extracts that and uses that as the private python for the IJulia stuff.

Question is, does that work for us? And where could we host that zip file? Right now it downloads from my public dropbox, which is obviously not a solution.

@tkelman, @stevengj, thoughts?

**Comment by tkelman** [comment]
We can put it anywhere, sourceforge or S3 would work, S3 would be more reliable but isn't completely free, and needs someone with upload permissions to do it. Not a big deal since these binaries look like they'll be smaller and change much less frequently than the Julia binaries we build continuously for a bunch of platforms already.

What exactly had to be done to repackage the installer? If it can't be scripted reliably at install-time on users' machines, can it at least be semi-reliably scripted or thoroughly documented for (less frequent) repackaging time on developers' machines?

**Comment by davidanthoff** [comment]
I looked into hosting binaries on sourceforge but just couldn't for the heck of it figure out a stable URL for a given file that would give me the download. Any pointer on that would be appreciated.

The repacking on a dev machine is very simple: install python 3.4 as a machine install, deselect all optional things in the installer, copy the two MSVC runtime dlls from systems32 into the python34 directory and zip everything up. Done. I'm 99% positive that this can be scripted as well. The bigger issue is that I like to do this whole exercise on a clean Win install with nothing else on it (so I do it on a VM), just to be sure that there are no old bits or anything like that.

**Comment by tkelman** [comment]
I've been putting a bunch of things up here https://sourceforge.net/projects/juliadeps-win/files/ - just add the filename to the end of that, and download using something that follows redirects and you're fine.

We should double-check the license terms for the msvc runtimes. They're generally redistributable, and I think we'd be okay here since Python isn't GPL, but it could be problematic if we eventually wanted to bundle this along with the GPL libraries in base Julia.

**Comment by davidanthoff** [comment]
I tried to use ``download`` from base, is that following redirects? If not, any better suggestion?

I have a VS license, so my understanding is that I can redistribute the MSVC runtime.

**Comment by davidanthoff** [comment]
I can't get the download from sourceforge to work with ``download`` from base, so for now I'm having the binary in my dropbox folder.

@tkelman and @stevengj can you try whether the current bits work for you? The instructions at the top should do it. Thanks.

**Comment by tkelman** [comment]
You can try the `download_cmd` from BinDeps. I have a powershell option in https://github.com/JuliaLang/BinDeps.jl/pull/124 but it needs someone else to test it and verify whether it's useful.

Is the copy of Python that you have packaged here 32 bit or 64? Do things need to match between IPython and IJulia there, or are there no shared-library connections involved here?

**Comment by stevengj** [comment]
Things don't need to match between Julia and IPython, because they only communicate over ZMQ.

**Comment by davidanthoff** [comment]
@tkelman This is 32 bit Python, and that should work with both 32 and 64 bit Julia.

**Comment by davidanthoff** [comment]
@tkelman I tried the ``download_cmd`` command from BinDeps, but I can't get it to work. I'm using

````
download_cmd("https://sourceforge.net/projects/minimalportablepython/files/python-3.4.2.zip", "python.zip")
````
But that doesn't download anything. Any pointers what I should be doing instead?

**Comment by tkelman** [comment]
`BinDeps.download_cmd` should return a command object - most likely ``` `curl -o python.zip -L https://sourceforge.net/projects/minimalportablepython/files/python-3.4.2.zip` ``` with the current version of BinDeps, so you'll need to call `run(BinDeps.download_cmd(...))` on it.

**Comment by tkelman** [comment]
It definitely looks like the `download` from Base, which uses `URLDownloadToFileW` on Windows, just isn't following redirects. Might be worth raising an issue with base Julia to figure out how to make it do that.

**Comment by davidanthoff** [comment]
Alright, it now downloads the python binaries from sourceforge, thanks for the help! I think this is ready for testing by others.

File: deps/build.jl [hunk 1]
```diff
@@ -5,94 +5,77 @@
...
-include("ipython.jl")
-const ipython, ipyvers = find_ipython()
+juliaprofiles = Array(String,0)
 
-if ipyvers < v"1.0.0-dev"
-    error("IPython 1.0 or later is required for IJulia, got $ipyvers instead")
-else
-    eprintln("Found IPython version $ipyvers ... ok.")
-end
+if @windows? true : false
```
tkelman: stylistically I think `@windows_only begin` would look better here

tkelman:
It looks like this works on 32 bit Julia, I'll try 64 bit later. Whatever is happening between `Successfully installed pip-6.0.8 setuptools-12.1` and the first `[ProfileCreate]` really badly needs some feedback to the user that it's doing something though, it takes a while and looks completely dead. Is `-qqq` the absolute quietest setting for pip? Can we change that to something that isn't completely silent?

And this probably doesn't need to download, extract, and set up python again if you call `Pkg.build("IJulia")` a second time and the desired version has already been downloaded, extracted, and set up properly... Would be good if there were an early success detection, "looks like things are already set up and working here" with perhaps a tag file or something that you could manually delete if you needed to force a complete reinstall.

deps/build.jl
```diff
@@ -5,94 +5,77 @@
...
 
-# add Julia kernel manager if we don't have one yet
-if VERSION >= v"0.3-"
-    binary_name = "julia"
-else
-    binary_name = "julia-basic"
+    pythonzipfilename = normpath(downloadsdir, "python-3.4.2.zip")
+    getpipfilename = normpath(downloadsdir,"get-pip.py")
+    run(download_cmd("https://sourceforge.net/projects/minimalportablepython/files/python-3.4.2.zip", "$pythonzipfilename"))
+    download("https://bootstrap.pypa.io/get-pip.py", "$getpipfilename")
```
tkelman: do we need this separately? don't the latest versions of python have ensurepip included?

deps/build.jl
davidanthoff: Done

deps/build.jl
davidanthoff: Good point, I now use ensurepip

davidanthoff:
> Is -qqq the absolute quietest setting for pip? Can we change that to something that isn't completely silent?

Done

> And this probably doesn't need to download, extract, and set up python again if you call Pkg.build("IJulia") a second time and the desired version has already been downloaded, extracted, and set up properly... Would be good if there were an early success detection, "looks like things are already set up and working here" with perhaps a tag file or something that you could manually delete if you needed to force a complete reinstall.

I added support for this. Essentially if a previous build.jl call was successful, then the next call will only do upgrade commands for pip and ipython for the existing private python installation. I still recreate the private ipython profile that lives in ``usr/.ijulia`` on every build.jl call. This is mainly to avoid another update mechanism we might need, namely one where the profile is updated.

deps/build.jl
```diff
@@ -5,94 +5,89 @@
...
-        eprintln("Creating $prof with $s = $val...")
-        open(p, "w") do f
-            print(f, """
-c = get_config()
-c.$s = $val
-""")
+        using BinDeps
+
+        if ispath(downloadsdir)
+            run(`cmd /C RD "$downloadsdir" /S /Q`)
```
tkelman: could these be `rm(downloadsdir, recursive=true)` instead of `cmd /C RD` ?

tkelman:
This is looking really good. I'll make a point to test this in combination with `Pkg.checkout("BinDeps")` since I just merged that PR that makes the default `download_cmd` use powershell for better results when behind proxies.

deps/build.jl
davidanthoff: Ha, I didn't find the ``recursive`` option when I originally looked. Now changed.

deps/build.jl
```diff
@@ -5,94 +5,91 @@
...
-const ipython, ipyvers = find_ipython()
+juliaprofiles = Array(String,0)
 
-if ipyvers < v"1.0.0-dev"
-    error("IPython 1.0 or later is required for IJulia, got $ipyvers instead")
-else
-    eprintln("Found IPython version $ipyvers ... ok.")
-end
+@windows_only begin
+    ipython_version_to_install = "2.4.1"
```
davidanthoff: One question is still whether we should hardcode the IPython version that is installed or just always get the latest. Right now it is hard coded. I'm worried that if we just get the latest version instead things might break. E.g., as soon as 3.0 is released, it would be used, and who knows whether we might need to adjust things. So I would vote to keep things as they are with a hardcoded version, but would like to hear other input as well. It does mean we would have to keep this line updated.

tkelman:
Markupsafe and tornado are invoking MSVC when they install on my machine. I hope they would continue to function correctly when MSVC is not installed...

davidanthoff:
I've tested this on a fresh Windows XP install throughout that certainly has no MSVC installed, and things seem to work. For tornado I know that it will try to compile things for speed, but if that fails for any reason it will still work. I assume (hope) Markupsafe follows a similar strategy.

tkelman:
Cool, I guess if the performance would suffer noticeably from not having the C extensions compiled, we can bug the developers of those packages to put up compiled binary wheels and that would resolve the performance difference for users who don't have MSVC installed?

davidanthoff:
Yes, I think that would work. I tried to test on XP again yesterday, but the julia ZMQ package failed to build. I'm traveling this week, but will try again on Fri.

tkelman:
> the julia ZMQ package failed to build

Presumably that's the same underlying issue as https://github.com/JuliaLang/ZMQ.jl/issues/69 and should go away after a restart of Julia. I wish people would write down error messages...

tkelman:
I think all this might need is a rebase and an eventual squash? The ZMQ install issue is fixed on nightlies, no restarting Julia necessary there.

davidanthoff:
There was a lot of change to ``build.jl`` on master this week so I very much doubt I can just merge master right now, will probably require a bit of work to sort that out.

stevengj:
Closing in this since the Conda alternative has been merged.

davidanthoff:
Excellent news!

PR #248, opened 2014-11-04
[WIP] try to fix custom.js

### LINKED ISSUE #246: Uncaught TypeError: undefined is not a function in custom.js
Friends,

I've just gone through a fresh install of IPython, Julia and IJulia on Ubuntu 14.04.1.  There seems to be a problem in custom.js in the IJulia profile that causes notebooks to not load and function.  I was able to duplicate this in multiple environments and reproduce it in each major browser on Windows, Linux and Mac OS X so it seems to be a legit problem. 
 
I also vetted it out with the IPython guys in their Gitter channel.  Switching from the IJulia profile to any IPython profile gets rid of the error.  Switching back to IJulia causes the problem to return.  I included my setup instructions for IPython, Julia and IJulia below.  Any guidance would be appreciated.

These are my installation instructions that I followed exactly for IPython and IJulia:  https://gist.github.com/dbartow/166114161e0c438600db

Here is a screenshot of the error happening in Chrome latest on Mac OS X.

screen shot 2014-11-03 at 9 58 31 am

Carreau:
Should fix #246 (partially) but work in progress. 

It is a due to a complex interaction between codemirror 4, that is now asynchronous, the kernelspecs of IPython master and a few other things. @dbartow might want to try that. 

Notebook will work but you won't have highliting though. 

https://github.com/ipython/ipython/pull/6850 should fix that.

### DIFF
--- a/deps/custom.js
+++ b/deps/custom.js
@@ -49,7 +49,9 @@
 // end of IPython unmodified version 
 
 
-$([IPython.events]).on('notebook_loaded.Notebook', function(){
+// loadmode param needed for requireMode later
+var do_config = function(events, IPython, CodeMirror, loadmode){
+events.on('notebook_loaded.Notebook', function(){
     // add here logic that should be run once per **notebook load**
     // (!= page load), like restarting a checkpoint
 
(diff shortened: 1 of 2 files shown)


Carreau:
Closing as IPython as evolved too much. and fix is simpler.

stevengj:
Okay, see #276.

Issue #258, opened 2015-01-09
bug in clear_history()


afniedermayer:
When calling clear_history(), I get this error message:

`delete!` has no method matching delete!(::Dict{Integer,UTF8String}, ::Int64, ::Nothing)
while loading In[3], in expression starting on line 1

 in clear_history at /home/andras/.julia/v0.3/IJulia/src/history.jl:21
 in clear_history at /home/andras/.julia/v0.3/IJulia/src/history.jl:31
 in clear_history at /home/andras/.julia/v0.3/IJulia/src/history.jl:35

stevengj:
Fixed by #259

PR #259, opened 2015-01-09
fix bug in clear_history

afniedermayer:
Fixes issue #258


stevengj:
Thanks!

Probably this bug was introduced in the aftermath of JuliaLang/julia#3439

Issue #31, opened 2013-08-01
handle Kernel Interrupt (SIGINT?)


stevengj:
In the notebook, there is a menu item "Kernel Interrupt", which I suspect sends a SIGINT to the kernel process.  It would be good to catch this as a keyboard interrupt, just as we do in the REPL.

(I seem to remember there was a function somewhere in the Julia REPL startup which installed the signal handler for this, so we would just need to call this function and catch the exception in appropriate places.)

PR #261, opened 2015-01-19
Prevent SIGINT from killing the kernel

darwindarak:
Clicking the Interrupt Kernel button is still causing the kernel to restart.  This PR implements the fix mentioned in #31 to match the change in how Julia handles interrupts (JuliaLang/julia@10c7d4e).

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -12,7 +12,11 @@ include("inline.jl")
 using IPythonDisplay
 pushdisplay(InlineDisplay())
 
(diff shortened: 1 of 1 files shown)


garrison:
This fixes the issue for me.

stevengj:
Thanks!  There needs to be a `VERSION > v"0.3"` check, however, since Julia 0.2 is still supported by IJulia.

darwindarak:
Ok, added the check!

stevengj:
Thanks!

PR #276, opened 2015-02-17
Fix custom.js for 3.0

Carreau:
Just fix some semi colon and wrap in a `if < 3` block to prevent exec on 3.0.

Technically custom.js is not needed on 3.0, but some people might upgrade IPython later, 
and the old custom js would just prevent notebook from loading. 

With this, custom js will just be no-op. 

With that, running `ipython notebook --profile=julia` with 3.0 should work, there are still some inconsistencies. 
(Kernel still named Python and logo is python, but kernel is actually julia)

Once kernelspec PR is ready (yes, still working on it), the above issues  should be fixed.

### DIFF
--- a/deps/custom.js
+++ b/deps/custom.js
@@ -58,47 +58,47 @@
 if (parseInt(IPython.version[0]) < 3) {
 
 
-$([IPython.events]).on('notebook_loaded.Notebook', function(){
-    "use strict";
-    // add here logic that should be run once per **notebook load**
-    // (!= page load), like restarting a checkpoint
+    $([IPython.events]).on('notebook_loaded.Notebook', function(){
+        "use strict";
+        // add here logic that should be run once per **notebook load**
+        // (!= page load), like restarting a checkpoint
 
-    var md = IPython.notebook.metadata;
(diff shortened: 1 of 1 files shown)


Carreau:
Use [`url?w=0`](https://github.com/JuliaLang/IJulia.jl/pull/276/files?w=0) to see the diff without the whitespace.

rgbkrk:
Looking great @Carreau, I'm excited!

stevengj:
Looks good, @Carreau, is this ready to merge?

Carreau:
From [this tweet](https://twitter.com/Mbussonn/status/567776207179157504) and after, I woudl say it works and have been tested by at least 2 people me included. So yes.

jiahao:
Works for me too! Tested on both 0.3.6+ and 0.4-dev

screen shot 2015-02-22 at 9 14 12 pm

Really liking the improved scrolling behavior.

Carreau:
> Really liking the improved scrolling behavior.

Have been using 3.0 for so long I don't remember what scrolling behavior ...

Issue #224, opened 2014-08-22
Support Jupyter (IPython 3) kernel spec


stevengj:
This should be a very simple matter of writing out a couple of extra files to the `.ipython` directory.

cc: @ellisonbg, @Carreau

stevengj:
@Carreau, since ipython#6537 was merged, is there anything else holding this up?

Carreau:
> @Carreau, since ipython#6537 was merged, is there anything else holding this up?

Me getting time to finis #210 among other. I should be definitively in berkeley in something like 48h.
probably will take a day or two to get office space and badges, and I resume my work.
(prey for delta airline to get power outlets, and I'll geek in the plane)

stevengj:
Cool, thanks!  So we are not waiting on any changes on the IPython side anymore.

Carreau:
Except for last minute [raptor](http://xkcd.com/292/), no.

jiahao:
It looks like both 0.3.6+ and 0.4-dev work quite nicely with 3.0.0rc1:

![screen shot 2015-02-22 at 9 14 12 pm](https://cloud.githubusercontent.com/assets/1732/6321962/c9721568-bad7-11e4-8993-320acdc5f4a5.png)

Really liking the improved scrolling behavior!

jiahao:
It looks like the only thing left to do is to fix this deprecation warning:

```
/Library/Python/2.7/site-packages/IPython/kernel/manager.py:95: UserWarning: Setting kernel_cmd is deprecated, use kernel_spec to start different kernels.
```

Otherwise it looks like all systems go.

Carreau:
Just don't use `--profile=julia` and you are good to go. 
(unless you get some other custom things in your profile)

jiahao:
I don't need `--profile=julia` anymore??!! (mind blown)

Carreau:
We do out best so that you do not have to:

![capture d ecran 2015-02-22 a 18 37 12](https://cloud.githubusercontent.com/assets/335567/6322075/f1ce54a6-bac1-11e4-9e76-0a3b2a0a22cc.png)

Still a bit of work, but all languages should be equal citizen soon.

jiahao:
I would really like to have both Julia 0.3 and 0.4 accessible from Jupyter. Is that possible now or is that still WIP?

Carreau:
[...]

PR #280, opened 2015-02-23
Allow multiple versions of Julia as IPython kernels

jiahao:
This change allows multiple Julia versions to be registered as
IPython/Jupyter 3.0 kernels.

You will get entries like

- Julia 0.3.6
- Julia 0.4.0-dev
- Julia 0.4.0-dev debug

depending on which Julia binaries have been used to build IJulia.jl.

To get multiple Julia versions available as IPython kernels, you will
need to rebuild IJulia from within each Julia binary.

screen shot 2015-02-22 at 10 47 57 pm

Ref: #224

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -144,7 +144,7 @@ copy_config("ijuliafavicon.ico",
 # notebook from loading.
 # todo: maybe do not copy if don't exist.
 # todo: maybe remove if user custom.js is identical to the one
-# shiped with IJulia ?
+# shipped with IJulia ?
 copy_config("custom.js", joinpath(juliaprof, "static", "custom"))
 
 # julia.js implements a CodeMirror mode for Julia syntax highlighting in the notebook.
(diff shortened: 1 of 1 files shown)


jakebolewski:
Why not allow the full VERSION in the spec, that way you could toggle between different -dev builds.

jiahao:
I was going by @Carreau's suggestion in [#224 (comment)](https://github.com/JuliaLang/IJulia.jl/issues/224#issuecomment-75482527) to use only the major version in `juliakspec` analogous to what they do for Python.

I also used `VERSION_STRING` ("0.4.0-dev") instead of `VERSION` ("0.4.0-dev+1234") in the `display_name`.

It appears that each time the version number bumps IPython will complain that the particular kernel spec no longer exists and wants to be pointed to the new version.

Carreau:
> It appears that each time the version number bumps IPython will complain that the particular kernel spec last used to open the notebook no longer exists and wants to be pointed to the new version.

That should be true only if there are many kernels that match the language 'julia'. 

If the exact match is not found ~~IPython~~ Jupyter will search a match on the language field, 
select it if there is only one match. Ask which one otherwise. We are ready if you have more suggstion 
for this behavior. 

Also IIUC, each time you build a version, the julia binary would be overwritten which would make the old kernelspec wrong/unusable.

stevengj:
Great, thanks!

PR #298, opened 2015-04-24
Remove space in kernel spec path

rgbkrk:
On IPython 3.1, the space was causing problems with loading the logo (even though the kernel spec itself gets loaded).

```
[W 20:07:18.060 NotebookApp] 404 GET /kernelspecs/julia%200.3/logo-64x64.png
```

Perhaps this bug exists in IPython notebook itself for kernelspecs with a space in the name.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,5 +1,3 @@
-# TODO: Buld IPython 1.0 dependency? (wait for release?)
-
 #######################################################################
 import JSON
 using Compat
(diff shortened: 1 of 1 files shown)


rgbkrk:
Ended up talking to @minrk of IPython/Jupyter, he had this to say:

> I think we are going to exclude spaces from valid kernelspec directory names, so we don't have to deal with URL escaping.

This fixes up the issue then.

stevengj:
It would probably be more readable if we replaced the space with a hyphen.

rgbkrk:
> It would probably be more readable if we replaced the space with a hyphen, though.

Done. I also removed a fairly outdated comment, now that IJulia is compatible with the current release of IPython and is right on par with the 4.0 release.

PR #314, opened 2015-06-05
use QUEUE(ROUTER) for heartbeat

minrk:
instead of FORWARDER(REP)

due to change in zmq_proxy in libzmq-4.0.6.

[more info](https://groups.google.com/forum/#!topic/jupyter/UM-ShiI_Mrs)


stevengj:
Thanks!

PR #315, opened 2015-06-08
Fix #305

### LINKED ISSUE #305: IJulia does not display same help as REPL for DomainError
It would be nice if the same help for DomainError is shown in IJulia as in the REPL. Bacause many newcommers to the language uses IJulia and do not understand the reasons behind the DomainError.
Output from IJulia:
```
in [1]: 1^(-10)
DomainError
while loading In[9], in expression starting on line 1

 in power_by_squaring at intfuncs.jl:60 (repeats 2 times)
```
In the REPL
```
julia> 1^(-10)
ERROR: DomainError
Cannot raise an integer x to a negative power -n. Make x a float by adding
a zero decimal (e.g. 2.0^-n instead of 2^-n), or write 1/x^n, float(x)^-n, or (x//1)^-n
 in power_by_squaring at intfuncs.jl:60 (repeats 2 times)
```
The main reason for this, is the help context is based on the backtrace for the `showerror` function for DomianErrors, see: https://github.com/JuliaLang/julia/blob/aa756622ea349cd28788509f32b23a05a42dbbb4/base/replutil.jl#L94-L110. The backtrace is not send to the `showerror` function at https://github.com/JuliaLang/IJulia.jl/blob/bcaa47203cd12cfe25d3fa7e0d5630499a5fa8d9/src/execute_request.jl#L71.

I would be happy to sent a PR but I do not know which way to proceed.
   * It could be to make the print of backtrace in `showerror(io::IO, ex::DomainError, bt)` at line https://github.com/JuliaLang/julia/blob/aa756622ea349cd28788509f32b23a05a42dbbb4/base/replutil.jl#L109 optional with a keyword and then special case the treatment of a `DomainError` in the `pyerr_content(e, msg::AbstractString="")`

Do any one have a better solution?

dhoegh:
Adapt IJulia to use the keyword argument added in JuliaLang/julia#11112 to fix #305

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -59,16 +59,18 @@ end
 
 # return the content of a pyerr message for exception e
 function pyerr_content(e, msg::AbstractString="")
+    bt = catch_backtrace()
     tb = map(utf8, @compat(split(sprint(Base.show_backtrace, 
                                         :execute_request_0x535c5df2, 
-                                        catch_backtrace(), 1:typemax(Int)),
+                                        bt, 1:typemax(Int)),
                                  "\n", keep=true)))
     if !isempty(tb) && ismatch(r"^\s*in\s+include_string\s+", tb[end])
         pop!(tb) # don't include include_string in backtrace
     end
     ename = string(typeof(e))
     evalue = try
-        sprint(showerror, e)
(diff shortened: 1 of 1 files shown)


stevengj:
Looks good, thanks!

PR #353, opened 2015-09-07
use `jupyter kernelspec install` when available

### LINKED ISSUE #351: Julia kernel update from 0.3.7 to 0.3.11
I uninstalled previous version of Julia 0.3.7 and installed 0.3.11.
After that I rebuilt and updated all packages including IJulia.

However, in the IPython notebook, I still only can see Julia 0.3.7 kernel and versioninfo() in IPython notebook shell says I'm using Julia 0.3.7. 

Could you let me know what I'm missing? Thanks!

minrk:
and use `ipython kernelspec install` for IPython 3, rather than attempting to reimplement the install process.

I didn't change the ipython2, ipython3 checks, but I'm not sure how one can get into a state where ipython2/3 exists but `ipython` does not.

closes #351

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -9,12 +9,16 @@ eprintln(x...) = println(STDERR, x...)
 ENV["PYTHONIOENCODING"] = "UTF-8"
 
 include("ipython.jl")
-const ipython, ipyvers = find_ipython()
+const command, cmdvers = find_kernelspec_cmd()
 
-if ipyvers < v"3.0"
-    error("IPython 3.0 or later is required for IJulia, got $ipyvers instead")
+if command == "jupyter"
+    eprintln("Found jupyter kernelspec version $cmdvers ... ok.")
 else
-    eprintln("Found IPython version $ipyvers ... ok.")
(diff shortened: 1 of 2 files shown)


minrk:
Changes you might want to make:

- make `--user` install optional (allow system-wide install)

stevengj:
Looks good, thanks!  I think the `--user` is right for the `Pkg.build` script; if someone wants a system-wide install, they should probably do that manually.

stevengj:
Anaconda has a program called `jupyter-kernelspec` .... is that related?

minrk:
`jupyter-kernelspec` *is* `jupyter kernelspec`. jupyter dispatches to subcommands like git, so `jupyter foo` calls `jupyter-foo`, etc.

stevengj:
The odd thing is that I got `ipython` and `jupyter-kernelspec` from `ipython-notebook`, but not `jupyter` unless I explicitly told conda to install `jupyter`.   Does `ipython kernelspec` dispatch to `jupyter-kernelspec`?

PR #355, opened 2015-09-11
Mention Conda.jl in readme

dhoegh:
@Luthaf made a great package enabling installation of Python + IPython + IJulia by a few lines of Julia code without any need for further setup to get running with IJulia.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,14 @@ formatted text, math, and multimedia in a single document
 
 First, [install IPython](http://ipython.org/install.html); you may
 also want some scientific-Python packages (SciPy and Matplotlib).
-The simplest way to do this on Mac and Windows is by [downloading the
-Anaconda package](http://continuum.io/downloads) and running its
-installer.  (Do *not* use Enthought Canopy/EPD.)
+The simplest way to do this on Mac and Windows is by using the
+Conda.jl package that installs the [Miniconda](http://conda.pydata.org/miniconda.html) Python distribution. Alternativly [Anaconda](http://continuum.io/downloads) can be installed. (Do *not* use Enthought Canopy/EPD.) To install Miniconda do:
+```
+Pkg.add("Conda")
+using Conda
+# Installing the notebook interface
+Conda.add("ipython-notebook") # important restart the julia terminal before running Pkg.add("IJulia")
+```
 
 * **Important**: on Windows, the Anaconda installer window gives options *Add Anaconda to the System Path* and also *Register Anaconda as default Python version of the system*.  Be sure to **check these boxes**.
 


Luthaf:
Conda.jl **should** be usable on Linux too, if you want to provide an unified interface. But it does not set environment values [anymore](https://github.com/Luthaf/Conda.jl/commit/0b395fef61509383094573a441b0f8cc26bdcc34#diff-b5a035973669724cfc435a3914eae6ebL86), which can be an issue if you try to call directly the binaries. This can be reverted if needed by IJulia!

tkelman:
This should probably be tested and automated everywhere (unless a new enough system ipython/jupyter is found) as a replacement for #211.

stevengj:
This sounds like a nice option!  I'm giving a Julia tutorial on Monday, so I will try this out on a few machines this weekend.

dhoegh:
I have used it with success on different windows computers last weekend. I have not tested it on osx as I have no such device.

README.md
```diff
@@ -14,9 +14,14 @@ formatted text, math, and multimedia in a single document
...
-The simplest way to do this on Mac and Windows is by [downloading the
-Anaconda package](http://continuum.io/downloads) and running its
-installer.  (Do *not* use Enthought Canopy/EPD.)
+The simplest way to do this on Mac and Windows is by using the
+Conda.jl package that installs the [Miniconda](http://conda.pydata.org/miniconda.html) Python distribution. Alternativly [Anaconda](http://continuum.io/downloads) can be installed. (Do *not* use Enthought Canopy/EPD.) To install Miniconda do:
+```
+Pkg.add("Conda")
+using Conda
+# Installing the notebook interface
+Conda.add("ipython-notebook") # important restart the julia terminal before running Pkg.add("IJulia")
```
stevengj: I'm a little confused by this.  `Pkg.build("IJulia")` currently looks for the `ipython` or `jupyter` commands in the global `PATH`.   Does `Conda.add` modify the global path in the shell `.profile` file or something?  That doesn't seem like great behavior for something that is supposed to be internal to Julia.

dhoegh: Conda.jl adds a full blown miniconda installation registret with the os through 'path'. It would be no different from manually installing miniconda and hence it is not private to Julia.

tkelman: Not added to the path any more: https://github.com/Luthaf/Conda.jl/commit/0b395fef61509383094573a441b0f8cc26bdcc34#diff-b5a035973669724cfc435a3914eae6ebL86

Would be better to refer to it by absolute path anyway.

stevengj: Yes, probably:

* The Conda package should export something like `Conda.binpath("jupyter")` to return the absolute path of a conda-installed executable.
* `IJulia/deps/jupyter.jl` should look in `JUPYTER` environment variable or similar(?) to allow an absolute path to be user-specified at `Pkg.build`-time (without modifying `PATH`)
* The Juptyer path should be output to a `const jupyter = "..."` in a `IJulia/deps/deps.jl` file that is generated during `Pkg.build`, and is `included` from `IJulia.jl`.

stevengj:
We might consider having IJulia:

* `REQUIRE` the `Conda` package
* Tell Conda to install ipython-notebook at `Pkg.build` time
* Install the kernelspec for *both* the `Conda` jupyter and any system-wide `jupyter` command that is found in the PATH.  (Or maybe they both use the same config directory so that this is redundant.)
* Use the Conda jupyter for the `IJulia.notebook()` function.

Although I would only want to do this if it seems to be working on Mac, Windows, and Linux.

README.md
dhoegh: @tkelman Conda installs miniconda in path on windows even after Luthaf/Conda.jl@0b395fe. I have made a pull request fixing this issue at https://github.com/Luthaf/Conda.jl/pull/7

dhoegh:
@stevengj sounds like a good plan. @Luthaf do you see any obstructions in realizing this?

dhoegh:
@stevengj by requiring Conda all IJulia users will get an miniconda installed even though they specify `JUPYTER` environment variable. Is this ok or could we work around it?

Luthaf:
> The Conda package should maybe export something like Conda.binpath("jupyter") to return the absolute path of a conda-installed executable.

This one can be difficult on Windows, because the binary will be in `$PREFIX/pkg/$pkg_name-$pkg_version/bin`. So would have to track the package name and version associated with the executable.

Maybe using an environment would do the trick.

> even though they specify `JUPYTER` environment variable. Is this ok or could we work around it?

Maybe the `build.jl` can detect if the `JUPYTER` variable is set before declaring dependencies.

stevengj:
@Luthaf, can't the `conda` command tell you the package name and version?

dhoegh:
By utilizing the second part of the regex in: https://github.com/Luthaf/Conda.jl/blob/4e39b971300c0519524a600c106000d06aa39c2f/src/Conda.jl#L104-L117 it should be able to tell the version.

Luthaf:
> @Luthaf, can't the conda command tell you the package name and version?

Only if the executable name matches the package name. This could be doable with meta-data search too.

stevengj:
In our case, however, we will know the package name, so it is fine to ask us to supply that, maybe in a `Conda.pkgpath(package::String)` function?

Luthaf:
> In our case, however, we will know the package name, so it is fine to ask us to supply that, maybe in a Conda.pkgpath(package::String) function?

Yes, this is doable. See https://github.com/Luthaf/Conda.jl/pull/8 for a first attempt doing this.

Issue #199, opened 2014-06-23
Installing IPython notebook and dependencies as part of IJulia setup.


johansigfrids:
Getting IPython setup can be a bit hairy, and is probably the trickiest bit I have to walk someone through when I want them to give Julia/IJulia a try. It would be great if this could happen automagically as part of installing IJulia. 

Getting it to work as part of the install could be hairy as well but maybe it could be made easier by packaging up IPython and dependecies into a binary with something like PyInstaller and then installing the resulting executable as a binary dependency.

htgoebel:
Please not: I just updates PyInstaller to support pyzmq 14.0+

tkelman:
@htgoebel thanks for the update. Does PyInstaller allow building 32 bit executables on 64 bit Windows at this time?

davidanthoff:
I've continued on my branch a little bit. Right now I'm using miniconda to install python, that seems to solve most of the problems I pointed out earlier. There are still a few problems on this, but I'm working on it and I should have some time for this now.

stevengj:
miniconda is something that should work on MacOS too, ideally, although I'm sure you have some Windows-specific stuff.

davidanthoff:
My code is Windows only for now, but I'm sure something very similar could be done on other platforms as well. My branch should be ready for other people to try out/test etc in a few days at most, I'm more or less done just want to do some more tests.

tkelman:
Done by #357!

PR #357, opened 2015-09-13
[WIP] Use Conda.jl if Ipython/jupyter is not found

dhoegh:
This is the first working version of an integrated IPython setup on all platforms.
* Make build.jl can detect a `JUPYTER` environment variable
* The Juptyer path should be output to a `const jupyter = "..."` in a IJulia/deps/deps.jl file that is generated during `Pkg.build`, and is included from IJulia.jl


REQUIRE
```diff
@@ -2,4 +2,4 @@ julia 0.3
 Nettle
 JSON 0.2
 ZMQ 0.1
-Compat 0.7
+Compat 0.7
```
stevengj: ?

deps/build.jl
```diff
@@ -2,14 +2,58 @@
 import JSON
 using Compat
 
+# remove deps.jl if it exists, in case build.jl fails
+isfile("deps.jl") && rm("deps.jl")
+
+if isdir(abspath(dirname(@__FILE__), "..", "..", "Conda"))
```
stevengj: Just add Conda to the `REQUIRE` file and assume it is installed.

deps/build.jl
```diff
@@ -2,14 +2,58 @@
...
 # print to stderr, since that is where Pkg prints its messages
 eprintln(x...) = println(STDERR, x...)
 
 # Make sure Python uses UTF-8 output for Unicode paths
 ENV["PYTHONIOENCODING"] = "UTF-8"
 
-include("jupyter.jl")
-const jupyter, jupyter_vers = find_jupyter()
+if Conda_installed
+    condapath=dirname(Conda.conda)
```
stevengj: Are you sure this is right?  On Windows, the `conda` command is in the `Scripts` directory.  Shouldn't it be `joinpath(Conda.PREFIX, "bin")`?

deps/build.jl
```diff
@@ -2,14 +2,58 @@
...
+    const jupyter, jupyter_vers = try
+        joinpath(condapath,"jupyter"),convert(VersionNumber, chomp(readall(`$(joinpath(condapath,"jupyter")) kernelspec --version`)))
+    catch e1
+        try
+            joinpath(condapath,"ipython"),convert(VersionNumber, chomp(readall(`$(joinpath(condapath,"ipython")) --version`)))
```
stevengj: What is the point of this?  Won't Jupyter always be installed, assuming you do `Conda.add("jupyter")`?

deps/build.jl
```diff
@@ -2,14 +2,58 @@
 import JSON
 using Compat
 
+# remove deps.jl if it exists, in case build.jl fails
+isfile("deps.jl") && rm("deps.jl")
+
+if isdir(abspath(dirname(@__FILE__), "..", "..", "Conda"))
+    Conda_installed = true
+    import Conda
+    Conda.add("ipython-notebook=3.1")
```
stevengj: Why are you specifying a version?

deps/build.jl
dhoegh: Ok. My thought was actually to detect if Conda.jl was installed else use system python. This is also the reason why I did not add Conda.jl to `REQUIRE`. If you think it is appropiat to add Conda as dependency i will do it.

deps/build.jl
stevengj: @Luthaf mentioned in https://github.com/JuliaLang/IJulia.jl/pull/355#issuecomment-139886213 that matters are more complicated on Windows.

dhoegh: It should be like this because of https://github.com/Luthaf/Conda.jl/blob/master/src/Conda.jl#L35-L36

stevengj: That is the location of the `conda` script, but is it the location of the `jupyter` program?  @Luthaf's comment indicates otherwise.

Luthaf: This would need the `Conda.bindir(package)` to be implemented first I think. I think the easier way to do this would be to ask callers of this function to do the executable <=> package linking by themselves.

deps/build.jl
stevengj: Probably we want to install `jupyter` (see JuliaLang/Interact.jl#73)

dhoegh:
I have updated the pull to a working condition. It do only work on master of Conda.jl. @Luthaf maybe you could tag a new version? While getting this to work I bumped into https://github.com/jupyter/notebook/issues/448. I have added a workaround for the issue.

REQUIRE
```diff
@@ -3,3 +3,4 @@ Nettle
 JSON 0.2
 ZMQ 0.1
 Compat 0.7
+Conda
```
stevengj: (please end the file with a newline)

deps/build.jl
```diff
@@ -78,4 +111,11 @@ copy_config("logo-32x32.png", juliakspec)
...
 eprintln("Installing julia kernelspec $spec_name")
-run(`$jupyter kernelspec install --replace --user $juliakspec`)
+run(`$jupyter-kernelspec install --replace --user $juliakspec`)
+
+open("deps.jl", "w") do f
+    print(f, """
+          const jupyter = "$(escape_string(jupyter))"
+          const jupyter_vers = $(repr(jupyter_vers))
+          """)
+end
```
stevengj: newline

deps/build.jl
```diff
@@ -78,4 +111,11 @@ copy_config("logo-32x32.png", juliakspec)
 copy_config("logo-64x64.png", juliakspec)
 
 eprintln("Installing julia kernelspec $spec_name")
-run(`$jupyter kernelspec install --replace --user $juliakspec`)
+run(`$jupyter-kernelspec install --replace --user $juliakspec`)
```
stevengj: I suppose you are doing this because of jupyter/notebook#448.

@minrk, will Jupyter always install `jupyter-kernelspec` into the same PATH as `jupyter`?  Are `jupyter-kernelspec` and `jupyter-notebook`  part of the public "API"?

I'm worried that at some point Jupyter may install the subsidiary programs like `jupyter-notebook` into a hidden internal path, or eliminate them entirely.

stevengj: Also, I don't think this will work if `$jupyter` is `ipython`, since I don't think the old IPython installed separate scripts?

Maybe the best thing is:

* Try `$jupyter kernelspec --version` and `$jupyter notebook --version` to see if they work.
* If they don't work, try `$jupyter-kernelspec` and `$jupyter-notebook`
* In `deps.jl`, define a `jupyter_notebook` constant to either `["$jupyter-notebook"]` or `[jupyter, "notebook"]`, and then run it in `notebook()` via  run(`$jupyter_notebook...`) to splat the arguments

dhoegh: Instead of circumventing jupyter/notebook#448 we could use IPython 3 by doing `Conda.add("ipython=3.*")` until jupyter/notebook#448 is closed.

minrk: The command should definitely be `jupyter kernelspec`, but this workaround should be fine while we work out what is going on with jupyter/notebook#448. We have no plans to install entrypoints to hidden locations any time soon.

Luthaf:
>  @Luthaf maybe you could tag a new version?

See JuliaLang/METADATA.jl#3488

dhoegh:
I have pushed a new version that fixes the workaround when IPython3 is used and the new lines.

deps/build.jl
```diff
@@ -78,4 +111,15 @@ copy_config("logo-32x32.png", juliakspec)
 copy_config("logo-64x64.png", juliakspec)
 
 eprintln("Installing julia kernelspec $spec_name")
-run(`$jupyter kernelspec install --replace --user $juliakspec`)
+if basename(jupyter) == "jupyter"
+    # Remove the commit that added this when https://github.com/jupyter/notebook/issues/448 is closed
+    run(`$jupyter-kernelspec install --replace --user $juliakspec`)
+else
+    run(`$jupyter-kernelspec install --replace --user $juliakspec`)
```
stevengj: Is this supposed to be `$jupyter kernelspec`?

deps/build.jl
```diff
@@ -1,15 +1,48 @@
...
+        try
+            "ipython2",convert(VersionNumber, chomp(readall(`ipython2 --version`)))
+        catch e3
+            try
+                "ipython3",convert(VersionNumber, chomp(readall(`ipython3 --version`)))
+            catch e4
+                try
+                    "ipython.bat",convert(VersionNumber, chomp(readall(`ipython.bat --version`)))
+                catch e5
+                    eprintln("No system wide Jupyter or IPython is found, got the following errors\n",
```
stevengj: system wide -> system-wide
is found -> was found
, got -> ; got
errors -> errors:

deps/build.jl
dhoegh: Yes it is.

deps/build.jl
```diff
@@ -1,15 +1,48 @@
...
 ENV["PYTHONIOENCODING"] = "UTF-8"
 
-include("jupyter.jl")
-const jupyter, jupyter_vers = find_jupyter()
+
+const jupyter, jupyter_vers = try
+    "jupyter",convert(VersionNumber, chomp(readall(`jupyter-kernelspec --version`)))
+catch e1
+    try
+        "ipython",convert(VersionNumber, chomp(readall(`ipython --version`)))
```
stevengj: I think these should check for `ipython` version 3 or later.

Maybe write:
```jl
function prog_version(prog)
    try
       return convert(VersionNumber, chomp(readall(`$prog --version`)))
    catch
       return v"0.0"
    end
end

jupyter = ""
jupyter_vers = v"0.0"
for p in ("jupyter","ipython","ipython2","ipython3","ipython.bat")
    v = prog_vers(p)
    if v >= v"3.0"
       jupyter = p
       jupyter_vers = v
       break
    end
end
if jupyter_vers < v"3.0"
    info("Could not find IPython or Jupyter 3.0 or later; installing via the Conda package.")
    Conda.add("jupyter")
    jupyter = joinpath(Conda.SCRIPTDIR,"jupyter")
    jupyter_vers = prog_vers(jupyter)
    jupyter_vers >= v"3.0" && error("failed to find $jupyter 3.0 or later")
end
```

dhoegh: If we intend to use `jupyter kernelspec` in the future when https://github.com/jupyter/notebook/issues/448 is fixed we should do:
```
function prog_version(prog, arg)
    try
       return convert(VersionNumber, chomp(readall(`$prog $arg --version`)))
    catch
       return v"0.0"
    end
end

jupyter = ""
jupyter_vers = v"0.0"
for (p arg) in (("jupyter", `kernelspec`),("ipython", ``),("ipython2", ``),("ipython3", ``),("ipython.bat", ``)) # for now ("jupyter", `kernelspec`) should be ("jupyter-kernelspec", ``) due to issue
    v = prog_vers(p, arg)
    if v >= v"3.0"
       jupyter = p
       jupyter_vers = v
       break
    end
end

Comment (stevengj) [inline deps/build.jl]: Both `jupyter` and `jupyter kernelspec` should return the same `--version`, so I think it should be sufficient to check for the former.

Comment (dhoegh) [inline deps/build.jl]: They do not seem to provide the same version number on my system:
```
C:\Users\Hoegh>jupyter --version
4.0.4

C:\Users\Hoegh>jupyter-kernelspec --version
4.0.0
```

Comment (stevengj) [inline deps/build.jl]: Hmm, same for me (MacOS with Anaconda).  @minrk, is this expected or is something wrong with Anaconda's Jupyter package?

In any case, I'm still not to worried as currently any `jupyter*` program should be fine (they all have versions > 3).

Comment (dhoegh) [inline deps/build.jl]: Ok. Then I will rebase to your version of the loop without the extra argument.

File: deps/build.jl [hunk 12]
```diff
@@ -78,4 +96,16 @@ copy_config("logo-32x32.png", juliakspec)
 copy_config("logo-64x64.png", juliakspec)
 
 eprintln("Installing julia kernelspec $spec_name")
-run(`$jupyter kernelspec install --replace --user $juliakspec`)
+if basename(jupyter) == "jupyter"
+    # Remove the commit that added this when https://github.com/jupyter/notebook/issues/448 is closed
```
stevengj: Just say `# Just use "$jupyter kernelspec" when jupyter/notebook#448 is resolved and widely available.`  I don't think I want to bother with a separate commit for this.

stevengj:
Looking good.  Can you squash the commits?

stevengj:
Oh nevermind, I guess we can leave it as two commits.

dhoegh:
Thank you for merging. The readme should probably be updated and close https://github.com/JuliaLang/IJulia.jl/issues/199 and https://github.com/JuliaLang/IJulia.jl/pull/211.

stevengj:
Yup, updating the README now.

tkelman:
:clap: :fireworks:

PR #370, opened 2015-10-08
create link to src/IJulia.gl

axsk:


stevengj:
Seems reasonable, thanks.

Issue #243, opened 2014-10-23
Writing more than 64 kb to stdout


phlpn:
I am calling something that writes more than 64 kb to stdout  from IJulia Notebook and the process stalls. Less than 64 kb works and doing the same thing on a worker process (or outside Ijulia notebook) works fine.

stevengj:
Re-opening this issue as it is not a Julia problem per se, it is an OS limitation for reading and writing a buffer device from the same thread.

The only solution would be to somehow instantiate a genuine OS thread that is responsible for reading the IO buffers; this is theoretically possible but will require some work (possibly at the C level).

Issue #347, opened 2015-08-30
Delayed output


rened:
Following up from https://groups.google.com/forum/#!topic/julia-users/nyb9qRJgzpo.

The issue is that the display of output to STDOUT is interrupted / delayed by computations. This test code

```jl
print(1:4); flush(STDOUT); svd(rand(1000,1000)); println("done"); 
```

prints "1:" and after the svd is done "4done", so flushing does not seem to help here.

This started to occur some time in the last weeks on Julia master, in 0.3.11 it works.

stevengj:
The issue is that the sending of `stdout` messages to Jupyter is handled by a separate green thread (it doesn't pre-empt `svd`), and exactly when that gets woken up is somewhat random.

Maybe IJulia could hook into `flush` to force it to wake up the stdout thread.

amellnik:
This may be an unrelated issue, but in 0.4 if you "run all cells" or similar you can end up with the output of anything that prints in the wrong cell:

![image](https://cloud.githubusercontent.com/assets/7132414/10113010/1d45bc56-6395-11e5-97e1-92ca7f2d1417.png)

stevengj:
@amellnik, that is #342.

amellnik:
Ah, apologies, I didn't look far enough back.

stevengj:
Actually, IJulia already hooks into `flush`.  It is just that it looks like the hooks broke in Julia 0.4.  Let me look into it further.

stevengj:
Okay, I fixed it so that the IJulia `flush` is actually being called.   However, at the time `flush` is called, after `1:` has printed but before the `4` has printed, `nb_available(STDOUT)` is returning `0`, so it thinks there is nothing to send.

@vtjnash, is this some kind of buffering?  How can the `flush` command get everything that has been written to the pipe at this point?

vtjnash:
The only way to get the kernel to give you everything on the pipe (with certainty), is to close the write end of the pipe and redirect_stdout to a new pipe for each new cell.

stevengj:
@vtjnash, that's not really acceptable for us, because there might be an `@async` task from a previous input cell still running and printing I/O to `STDOUT`.

So, this is really a kernel thing and there's nothing we can do about it?

stevengj:
Can we get the kernel to treat the `STDOUT` pipe as a pseudoterminal, so that it won't buffer I/O?

vtjnash:
[...]

Issue #327, opened 2015-07-02
test suite


stevengj:
As mentioned by @minrk in https://github.com/JuliaLang/IJulia.jl/pull/325#issuecomment-117816611, there is a new https://github.com/jupyter/jupyter_kernel_test repo for testing language kernels.  It would be good to use this at some point.

However, since it currently requires git `master` of IPython (3.2 is not good enough), it's probably not practical for us to use (even as an optional thing, or on Travis only) until a future release of IPython and jupyter.  (Some other required modules are ipython/traitlets, ipython/ipython_genutils, jupyter/jupyter_core, jupyter/jupyter_client, but it's possible to install these locally from git in the `tests` directory)

stevengj:
FYI, here was my first draft at a runtests.jl file: https://gist.github.com/stevengj/7d6af1bf87433ec9b805

It fails on my machine with `ImportError: No module named paths` because my IPython (3.2) is too old.

PR #392, opened 2015-11-24
RFC: Numerous IO related fixes

### LINKED ISSUE #238: stdout doesn't flush correctly
screen shot 2014-10-16 at 1 48 51 am

```
ccall((:printf, "libc"), Int, (Ptr{Uint8},), "abc");
```

It happens when I select "run all". It works fine if I only run that cell. It may be also related to #172.

Edit: 

It also seems that flushing doesn't help.

screen shot 2014-10-16 at 1 49 40 am

### LINKED ISSUE #342: stream messages come in irregular chunks
There is something unusual in the way IJulia is responding to the `whos()` command. Usually, in a fresh environment the output should be something like this:

```
julia> whos()
Base                          Module
Core                          Module
Main                          Module
ans                           Nothing
```

However, using the kernel directly, it outputs a couple of stream messages. The first message only shows one of the variables or modules in the environment and the second message outputs the rest. For example:

First message:

```
 'text': u'Base                          Module'},
```

and the second:

```
  'text': u'\nCompat                        Module\nCore                          Module\nIJulia
Module\nIPythonDisplay                Module\nJSON                          
Module\nMain                          Module\nNettle                        Module\nZMQ
Module\n'},
```

This is the complete session:

```
In [10]: kc.execute('whos()')
Out[10]: 'f118c20a-b199-4263-b133-fc801293b591'

In [11]: kc.get_iopub_msg(timeout=1)
Out[11]:
{'buffers': [],
 'content': {u'execution_state': u'busy'},
 'header': {u'msg_id': u'd00a1991-1c78-40d1-84cb-c88a830478d5',
  u'msg_type': u'status',
  u'session': u'????',
  u'username': u'jlkernel',
  'version': '5.0'},
 'metadata': {},
 'msg_id': u'd00a1991-1c78-40d1-84cb-c88a830478d5',
 'msg_type': u'status',
 'parent_header': {u'date': datetime.datetime(2015, 8, 22, 15, 46, 15, 447034),
  u'msg_id': u'f118c20a-b199-4263-b133-fc801293b591',
  u'msg_type': u'execute_request',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user'}}

In [12]: kc.get_iopub_msg(timeout=1)
Out[12]:
{'buffers': [],
 'content': {u'code': u'whos()', u'execution_count': 1},
 'header': {u'msg_id': u'28393ec3-1d4a-4d6f-b9fc-1a9f1e040809',
  u'msg_type': 'execute_input',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user',
  'version': '5.0'},
 'metadata': {},
 'msg_id': u'28393ec3-1d4a-4d6f-b9fc-1a9f1e040809',
 'msg_type': 'execute_input',
 'parent_header': {u'date': datetime.datetime(2015, 8, 22, 15, 46, 15, 447034),
  u'msg_id': u'f118c20a-b199-4263-b133-fc801293b591',
  u'msg_type': u'execute_request',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user'}}

In [13]: kc.get_iopub_msg(timeout=1)
Out[13]:
{'buffers': [],
 'content': {u'name': u'stdout',
  'text': u'Base                          Module'},
 'header': {u'msg_id': u'99c055ad-94be-405e-983e-f8237785b653',
  u'msg_type': u'stream',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user',
  'version': '5.0'},
 'metadata': {},
 'msg_id': u'99c055ad-94be-405e-983e-f8237785b653',
 'msg_type': u'stream',
 'parent_header': {u'date': datetime.datetime(2015, 8, 22, 15, 46, 15, 447034),
  u'msg_id': u'f118c20a-b199-4263-b133-fc801293b591',
  u'msg_type': u'execute_request',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user'}}

In [14]: kc.get_iopub_msg(timeout=1)
Out[14]:
{'buffers': [],
 'content': {u'name': u'stdout',
  'text': u'\nCompat                        Module\nCore                          Module\nIJulia                        Module\nIPythonDisplay                Module\nJSON                          Module\nMain                          Module\nNettle                        Module\nZMQ                           Module\n'},
 'header': {u'msg_id': u'265c7a5a-bf9e-4dda-952a-66c4c7c22305',
  u'msg_type': u'stream',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user',
  'version': '5.0'},
 'metadata': {},
 'msg_id': u'265c7a5a-bf9e-4dda-952a-66c4c7c22305',
 'msg_type': u'stream',
 'parent_header': {u'date': datetime.datetime(2015, 8, 22, 15, 46, 15, 447034),
  u'msg_id': u'f118c20a-b199-4263-b133-fc801293b591',
  u'msg_type': u'execute_request',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user'}}

In [15]: kc.get_iopub_msg(timeout=1)
Out[15]:
{'buffers': [],
 'content': {u'execution_state': u'idle'},
 'header': {u'msg_id': u'6aa2944a-e315-48dc-9078-e2c70a666836',
  u'msg_type': u'status',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'jlkernel',
  'version': '5.0'},
 'metadata': {},
 'msg_id': u'6aa2944a-e315-48dc-9078-e2c70a666836',
 'msg_type': u'status',
 'parent_header': {u'date': datetime.datetime(2015, 8, 22, 15, 46, 15, 447034),
  u'msg_id': u'f118c20a-b199-4263-b133-fc801293b591',
  u'msg_type': u'execute_request',
  u'session': u'592fe3e1-eeed-4094-b61d-878ee6023940',
  u'username': u'user'}}
```

I'm not sure why IJulia is using stream messages. As far as I know, other kernels (IPython, IR) are using `msg_type: execute_result` instead. Is this a bug?

### LINKED ISSUE #372: Large amounts of output leads to uninterruptable kernel hang
Hi, love IJulia, brilliant work. 

Currently running into a bug, I've managed to whittle reproduction down to this:

    for i=1:100000
        @show i
    end

Putting that in a notebook leads to the kernel "hanging" (shows the * in the notebook) and being uninterruptable (via the stop button), so every time I output "too much" I need to restart the kernel (restarting works fine, as does saving, while this is happening)

ijulia bug

That's the picture after scrolling to the bottom of the output.

With verbose = true, the last message is the output up to the number that the notebook shows:

    SENDING IPython Msg [ idents stdout ] {
      header = Dict("msg_id"=>"69ba5b82-bf10-43cf-8448-a41f6543dac0","msg_type"=>"stream","username"=>"username","version"=>"5.0","session"=>"C4F03FFDAA12480188C988F545458EF8"),
      metadata = Dict{AbstractString,Any}(),
      content = Dict("name"=>"stdout","text"=>"i = 19424\ni = 19425\ni = 19426\ni = 19427\n
    ...
    ni = 25787\ni = 25788\ni = 25789")
    }
    [I 00:10:12.219 NotebookApp] Saving file at /bugs.ipynb

I'm on IJulia master (and have run `Pkg.build("IJulia")` since getting the latest julia and `Pkg.checkout("IJulia")`), OSX 10.9, julia 0.4.0, jupyter 4.0.6 (installed via pip). Presumably this is related to #347 but thought I'd post separately, since the symptoms seem slightly more severe. Cheers.

JobJob:
- Fixes #372 by removing the `sleep(0.1)` in watch_stream (that was there to slow the rate of messages sent) and instead continually reading the `read_stdout` and `read_stderr` streams (whose buffers are limited to 64KB on OSX/Linux, 4KB Windows?) and adding data read into our own `IOBuffer`s (which are essentially limited only by memory size AFAIK)
    - The rate of sending is controlled by a `stream_interval` parameter; a stream message is sent at most once every `stream_interval` seconds (currently 0.1). The exception to this is if the buffer has reached the size specified in `max_size` (currently 10KB), and will then be sent immediately. This is to avoid overly large stream messages being sent. This mechanism also fixes #342.
- Fixes #238 #347
    - the ideas outlined here (https://github.com/JuliaLang/IJulia.jl/issues/347#issuecomment-144505862 and https://github.com/JuliaLang/IJulia.jl/issues/347#issuecomment-144605024) were implemented (see oslibuv_flush() in stdio.jl):
        - many thanks to @vtjnash for those suggestions
        - n.b. without the changes outlined above, these don't work, chiefly due to the `sleep(0.1)` in the existing watch stream implementation.
- Also fixes using `?keyword` (e.g. `?try`)
- Added timestamps to the debug logging and the task which the `vprintln` call is made from
- Notes:
    - Needs testing on Windows (and Linux for that matter, I'm on OSX)
    - Apologies for the occasional weird whitespace changes at the end of lines my editor is adding, not sure what's going on there. (n.b: you can add `?w=1` to any diff url on github to disable whitespace only changes)
    - We should discuss how we might add tests to the package

### DIFF
--- a/src/stdio.jl
+++ b/src/stdio.jl
@@ -4,13 +4,22 @@
 
 # logging in verbose mode goes to original stdio streams.  Use macros
 # so that we do not even evaluate the arguments in no-verbose modes
+
+function get_log_preface()
+    t = now()
+    taskname = get(task_local_storage(), :IJulia_task, "")
+    @sprintf("%02d:%02d:%02d(%s): ", Dates.hour(t),Dates.minute(t),Dates.second(t),taskname)
+end
+
+
 macro vprintln(x...)
     quote
         if verbose::Bool
-            println(orig_STDOUT, $(x...))
+            println(orig_STDOUT, get_log_preface(), $(x...))
         end
     end
 end
+
 macro verror_show(e, bt)
     quote
         if verbose::Bool
@@ -19,27 +28,31 @@ macro verror_show(e, bt)
     end
 end
 
-function send_stream(rd::IO, name::AbstractString)
-    nb = nb_available(rd)
-    if nb > 0
-        d = readbytes(rd, nb)
-        s = try
-            bytestring(d)
-        catch
-            # FIXME: what should we do here?
-            string("<ERROR: invalid UTF8 data ", d, ">")
-        end
-        send_ipython(publish,
-                     msg_pub(execute_msg, "stream",
-                             @compat Dict("name" => name, "text" => s)))
-    end
-end
+#name=>iobuffer for each stream ("stdout","stderr") so they can be sent in flush
+const bufs = Dict{ASCIIString,IOBuffer}()
+const stream_interval = 0.1
+const max_bytes = 10*1024
 
+"""Continually read from (size limited) Libuv/OS buffer into an (effectively unlimited) `IObuffer`
+to avoid problems when the Libuv/OS buffer gets full (https://github.com/JuliaLang/julia/issues/8789).
+Send data immediately when buffer contains more than `max_bytes` bytes. Otherwise, if data is available
+it will be sent every `stream_interval` seconds (see the Timers set up in watch_stdio)"""
 function watch_stream(rd::IO, name::AbstractString)
+    task_local_storage(:IJulia_task, "read $name task")
     try
+        buf = IOBuffer()
+        bufs[name] = buf
         while !eof(rd) # blocks until something is available
-            send_stream(rd, name)
-            sleep(0.1) # a little delay to accumulate output
+            nb = nb_available(rd)
+            if nb > 0
+                write(buf, readbytes(rd, nb))
+            end
+            if buf.size > 0
+                if buf.size >= max_bytes
+                    #send immediately
+                    send_stream(name)
+                end
+            end
         end
     catch e
         # the IPython manager may send us a SIGINT if the user
@@ -50,6 +63,62 @@ function watch_stream(rd::IO, name::AbstractString)
             rethrow()
         end
     end
+end
+
+function send_stdio(name)
+    if verbose::Bool && !haskey(task_local_storage(), :IJulia_task)
+        task_local_storage(:IJulia_task, "send $name task")
+    end
+    send_stream(name)
+end
+
+send_stdout(t::Timer) = send_stdio("stdout")
+send_stderr(t::Timer) = send_stdio("stderr")
+
+function send_stream(name::AbstractString)
+    buf = bufs[name]
+    if buf.size > 0
+        d = takebuf_array(buf)
+        n = num_utf8_trailing(d)
+        dextra = d[end-(n-1):end]
+        resize!(d, length(d) - n)
+        s = UTF8String(d)
+        if isvalid(s)
+            write(buf, dextra) # assume that the rest of the string will be written later
+            length(d) == 0 && return
+        else
+            # fallback: base64-encode non-UTF8 binary data
+            sbuf = IOBuffer()
+            print(sbuf, "base64 binary data: ")
+            b64 = Base64EncodePipe(sbuf)
+            write(b64, d)
+            write(b64, dextra)
+            close(b64)
+            print(sbuf, '\n')
+            s = takebuf_string(sbuf)
+        end
+        send_ipython(publish,
+             msg_pub(execute_msg, "stream",
+                     @compat Dict("name" => name, "text" => s)))
+    end
+end
+
+"""
+If `d` ends with an incomplete UTF8-encoded character, return the number of trailing incomplete bytes.
+Otherwise, return `0`.
+"""
+function num_utf8_trailing(d::Vector{UInt8})
+    i = length(d)
+    # find last non-continuation byte in d:
+    while i >= 1 && ((d[i] & 0xc0) == 0x80)
+        i -= 1
+    end
+    i < 1 && return 0
+    c = d[i]
+    # compute number of expected UTF-8 bytes starting at i:
+    n = c <= 0x7f ? 1 : c < 0xe0 ? 2 : c < 0xf0 ? 3 : 4
+    nend = length(d) + 1 - i # num bytes from i to end
+    return nend == n ? 0 : nend
 end
 
 # this is hacky: we overload some of the I/O functions on pipe endpoints
@@ -87,20 +156,40 @@ function readline(io::StdioPipe)
 end
 
 function watch_stdio()
-    @async watch_stream(read_stdout, "stdout")
+    task_local_storage(:IJulia_task, "init task")
+    read_task = @async watch_stream(read_stdout, "stdout")
+    #send STDOUT stream msgs every stream_interval secs (if there is output to send)
+    Timer(send_stdout, stream_interval, stream_interval)
     if capture_stderr
-        @async watch_stream(read_stderr, "stderr")
+        readerr_task = @async watch_stream(read_stderr, "stderr")
+        #send STDERR stream msgs every stream_interval secs (if there is output to send)
+        Timer(send_stderr, stream_interval, stream_interval)
     end
+end
+
+function flush_all()
+    flush_cstdio() # flush writes to stdout/stderr by external C code
+    flush(STDOUT)
+    flush(STDERR)
+end
+
+function oslibuv_flush()
+    #refs: https://github.com/JuliaLang/IJulia.jl/issues/347#issuecomment-144505862
+    #      https://github.com/JuliaLang/IJulia.jl/issues/347#issuecomment-144605024
+    @windows_only ccall(:SwitchToThread, stdcall, Void, ())
+    yield()
+    yield()
 end
 
 import Base.flush
 function flush(io::StdioPipe)
     invoke(flush, (super(StdioPipe),), io)
-    # send any available bytes to IPython (don't use readavailable,
-    # since we don't want to block).
     if io == STDOUT
-        send_stream(read_stdout, "stdout")
+        oslibuv_flush()
+        send_stream("stdout")
     elseif io == STDERR
-        send_stream(read_stderr, "stderr")
+        oslibuv_flush()
+        send_stream("stderr")
     end
 end
+
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -133,6 +133,7 @@ include("handlers.jl")
 include("heartbeat.jl")
 
 function eventloop(socket)
+    task_local_storage(:IJulia_task, "write task")
     try
         while true
             msg = recv_ipython(socket)
@@ -147,7 +148,7 @@ function eventloop(socket)
                 if !isa(e, InterruptException)
                     content = error_content(e, msg="KERNEL EXCEPTION")
                     map(s -> println(orig_STDERR, s), content["traceback"])
-                    send_ipython(publish, 
+                    send_ipython(publish,
                                  execute_msg == nothing ?
                                  Msg([ "error" ],
                                      @compat(Dict("msg_id" => uuid4(),
@@ -156,7 +157,7 @@ function eventloop(socket)
                                                   "msg_type" => "error",
                                                   "version" => "5.0")),
                                      content) :
-                                 msg_pub(execute_msg, "error", content)) 
+                                 msg_pub(execute_msg, "error", content))
                 end
             finally
                 send_status("idle", msg.header)
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -23,7 +23,7 @@ metadata(x) = Dict()
 # return a AbstractString=>AbstractString dictionary of mimetype=>data
 # for passing to Jupyter display_data and execute_result messages.
 function display_dict(x)
-    data = @compat Dict{ASCIIString,ByteString}("text/plain" => 
+    data = @compat Dict{ASCIIString,ByteString}("text/plain" =>
                                         sprint(writemime, "text/plain", x))
     if mimewritable(image_svg, x)
         data[string(image_svg)] = stringmime(image_svg, x)
@@ -74,7 +74,7 @@ end
 function error_content(e; backtrace_top::Symbol=:execute_request_0x535c5df2, msg::AbstractString="")
     bt = catch_backtrace()
     tb = map(utf8, @compat(split(sprint(show_bt,
-                                        backtrace_top, 
+                                        backtrace_top,
                                         bt, 1:typemax(Int)),
                                  "\n", keep=true)))
     if !isempty(tb) && ismatch(r"^\s*in\s+include_string\s+", tb[end])
@@ -131,8 +131,11 @@ function helpcode(code::AbstractString)
     else # new Base.Docs.@repl macro from julia@08663d4bb05c5b8805a57f46f4feacb07c7f2564
         code_ = strip(code)
         # as in base/REPL.jl, special-case keywords so that they parse
-        return "Base.Docs.@repl " * (haskey(Docs.keywords, symbol(code_)) ?
-                                     ":"*code_ : code_)
+        if(haskey(Docs.keywords, symbol(code_)))
+          return "eval(:(Base.Docs.@repl \$(symbol(\"$code_\"))))"
+        else
+          return "Base.Docs.@repl $code_"
+        end
     end
 end
 
@@ -148,20 +151,20 @@ function execute_request_0x535c5df2(socket, msg)
 
     if !silent
         _n += 1
-        send_ipython(publish, 
+        send_ipython(publish,
                      msg_pub(msg, "execute_input",
                              @compat Dict("execution_count" => _n,
                                           "code" => code)))
     end
-    
+
     silent = silent || ismatch(r";\s*$", code)
     if store_history
         In[_n] = code
     end
 
     # "; ..." cells are interpreted as shell commands for run
-    code = replace(code, r"^\s*;.*$", 
-                   m -> string(replace(m, r"^\s*;", "Base.repl_cmd(`"), 
+    code = replace(code, r"^\s*;.*$",
+                   m -> string(replace(m, r"^\s*;", "Base.repl_cmd(`"),
                                "`)"), 0)
 
     # a cell beginning with "? ..." is interpreted as a help request
@@ -170,17 +173,20 @@ function execute_request_0x535c5df2(socket, msg)
         code = helpcode(hcode)
     end
 
-    try 
+    try
         for hook in preexecute_hooks
             hook()
         end
+
+        #run the code!
         ans = result = include_string(code, "In[$_n]")
+
         if silent
             result = nothing
         elseif result != nothing
             if store_history
                 if result != Out # workaround for Julia #3066
-                    Out[_n] = result 
+                    Out[_n] = result
                 end
             end
         end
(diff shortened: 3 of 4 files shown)


stevengj:
Removing the `sleep` won't fix #372, I think; the only way to fix that is to read from a separate OS thread (not a green thread / task).

JobJob:
Yeah it doesn't fix #243 (which will need a new thread - waiting for that stuff to settle in Julia 5 before trying something), but it does fix output coming at too fast a rate, such that the buffer gets to 64kb, which is what was happening to me in #372

stevengj:
Overall, it looks very reasonable.  The `throttle` function seems like overkill here, though.  Wouldn't it be sufficient to have something like:
```
elseif time() > prev_time + stream_interval
    prev_time = time()
    send_stream(...)
end
```
in `watch_stream`?

JobJob:
I think you might be right, I was planning to possibly separate the throttle out, possibly with another couple of timing related functions, e.g. a debounce (http://underscorejs.org/#debounce), into a package of their own, so wrote something more general than is perhaps needed. An earlier version had something like that and it made the code a bit messier (but perhaps that was because I was trying to reset the timer at the start of execution - which didn't fix (the contentious contentions of) #342 - the timer needs to be started from when the first output might have been else the delay is too long and the first bytes read will be sent). 

I'll see if I can easily make it simpler tomorrow (off to bed soon).

Not sure if I've thanked you for making the package yet - great form.

JobJob:
Oh now I remember. The issue with an implementation like that is that say the `eof` wakes up with some data, it's read, but not sufficient time has passed since the last `send_stream` (say 0.05 seconds - with `stream_interval` say 0.1) so you would decide not to send immediately, but `watch_stream` then returns to the `eof` and blocks until further output, and the message isn't sent. 

So what you'd need to do is set a Timer to schedule a `send_stream` at 0.05 seconds from now. But, if more output comes in the next 0.05 seconds, you need to check if there's a timer already set and set one up if not, which is basically what `throttle` does, but abstracts away that complexity from the code that uses it.

JobJob:
Just found out you can add `?w=1` to the end of any diff url on github to hide trailing whitespace changes: https://github.com/JuliaLang/IJulia.jl/pull/392/files?w=1

JobJob:
Ok, pushed a much simpler solution. Two timers set up in `watch_stdio` that just continuously call `send_stream` for each stream (see latest commit). Note, `send_stream` won't send if the buffer is empty.

As long as the flush works correctly after each cell is done (which it has to else you get https://github.com/JuliaLang/IJulia.jl/issues/342#issuecomment-147856782) this should be fine. The one possible drawback is that the first message after the cell gets executed will be smaller than average - since where output starts within the `stream_interval` cycle will be random, on average there will only be `stream_interval/2` secs to accumulate output before the first stream message is sent and occasionally you will get unlucky and only a very small amount of output will be sent in the first message and then you'll have to wait a whole `stream_interval` seconds for the next bit :) 

Hopefully this will be acceptable regarding this:  https://github.com/JuliaLang/IJulia.jl/issues/342#issuecomment-133761404. So far for me the IO seems a little snappier in general after these changes, so let's see, maybe there was some other issue that was exacerbating the problem there.

src/stdio.jl
```diff
@@ -4,13 +4,42 @@
...
+    ms = t - floor(t) #time() returns utc, now() returns local time
+    dt = Dates.unix2datetime(floor(t+utcoffset))
+    ts = "$(string(dt)[12:end]).$(string(ms)[3:end])"
+end
+
+function get_log_preface()
+    ts = get_log_ts()
+    task = current_task()
+    taskname =
+        if isdefined(IJulia, :tasks)
```
stevengj: Seems like it would be cleaner to use `task_local_storage`.  Have each task call `task_local_storage(:IJulia_task, "myname")` and then in `get_log_preface` use `get(task_local_storage(), :IJulia_task, "")`

src/stdio.jl
```diff
@@ -29,17 +59,37 @@ function send_stream(rd::IO, name::AbstractString)
             # FIXME: what should we do here?
             string("<ERROR: invalid UTF8 data ", d, ">")
         end
-        send_ipython(publish,
-                     msg_pub(execute_msg, "stream",
-                             @compat Dict("name" => name, "text" => s)))
+    else
+        s = ""
     end
+    s, nb
```
stevengj: It seems like this should just return `d` now.  There's no reason to convert it into a string if we are just going to write the bytes to a buffer in `watch_stream`

src/stdio.jl
```diff
@@ -52,6 +102,18 @@ function watch_stream(rd::IO, name::AbstractString)
     end
 end
 
+send_stdout(t::Timer) = send_stream("stdout")
+send_stderr(t::Timer) = send_stream("stderr")
+
+function send_stream(name::AbstractString)
+    if bufs[name].size > 0
+        s = takebuf_string(bufs[name])
```
stevengj: Should we check for an invalid UTF8 string here (ala the `try`/`catch` in `read_stream`)?  If there is a partially encoded UTF8 character at the end of the buffer, which I suppose could theoretically happen although I've never seen it, it would be good to omit it from `s` and push those bytes back into the buffer (in the hope that the rest of the encoding will be pushed later).

src/stdio.jl
JobJob: Didn't know about task_local_storage, sounds perfect.

src/stdio.jl
JobJob: Good point.

src/stdio.jl
JobJob: Yes, that sounds like a good idea.

JobJob: bathbytes = convert(Vector{UInt8}, "🛀")
    write(STDOUT, bathbytes[1:2])
    flush(STDOUT)
    sleep(2.0)
    write(STDOUT, bathbytes[3:4])

causes a pretty good error without the change you recommended.

JobJob:
Fixed those things, cheers.

src/stdio.jl
```diff
@@ -4,13 +4,34 @@
...
+    dt = Dates.unix2datetime(floor(t+utcoffset))
+    msstr = rpad(string(ms)[3:end], precision, "0")
+    ts = "$(string(dt)[12:end]).$msstr"
+end
+
+function get_log_preface()
+    ts = get_log_ts()
+    taskname = get(task_local_storage(), :IJulia_task, "")
+    "$ts($taskname): "
+end
```
stevengj: This seems overly complicated.  Why not just:
```jl
function get_log_preface()
    t = now()
    taskname = get(task_local_storage(), :IJulia_task, "")
    string(Dates.hour(t),':',Dates.minute(t),':',Dates.second(t),'(',taskname,')')
end
```

JobJob:
FYI: I've started working on a branch where output from long running tasks will be appended to each cell. See: https://github.com/JobJob/IJulia.jl/tree/per-cell-stream
This test case seems to work currently with that branch
Cell 1:

    t = @schedule (cur_std = STDOUT; sleep(2); println(cur_std, "SoDo"); sleep(2); println(cur_std, "SoPa"))

Cell 2:

    println("1")

i.e. it prints "SoDo SoPa" in cell 1.
One issue is I have to not send the idle message. Not sure really how to handle that, maybe it's correct for the cell to not send the idle, or maybe there should be a call, e.g. `IJulia.enable_long_running_cells()`, that stops idles from being sent.
What are your thoughts?

stevengj:
I think an idle message should be sent as soon as IJulia is ready to execute another cell, regardless of what is running in the background.

src/stdio.jl
```diff
@@ -19,27 +40,38 @@ macro verror_show(e, bt)
...
-        catch
-            # FIXME: what should we do here?
-            string("<ERROR: invalid UTF8 data ", d, ">")
-        end
-        send_ipython(publish,
-                     msg_pub(execute_msg, "stream",
-                             @compat Dict("name" => name, "text" => s)))
-    end
+    d = nb > 0 ? readbytes(rd, nb) : UInt8[]
+    d, nb
```
stevengj: Why do we need to return `nb`?  Isn't this equivalent to `length(d)`?

src/stdio.jl
```diff
@@ -19,27 +40,38 @@ macro verror_show(e, bt)
...
 end
 
+#name=>iobuffer for each stream ("stdout","stderr") so they can be sent in flush
+const bufs = Dict{AbstractString,IOBuffer}()
```
stevengj: `ASCIIString`, not `AbstractString`.

src/stdio.jl
JobJob: `now()` doesn't return milliseconds. `time()` gives UTC time to high precision. I needed high precision to get visibility into some timing issues with async IO. The formatting is nice, and you get it in local time.

JobJob:
When you send the idle Jupyter won't display any more output AFAICT

src/stdio.jl
```diff
@@ -52,6 +84,43 @@ function watch_stream(rd::IO, name::AbstractString)
...
+    end
+end
+
+function send_stdout(t::Timer)
+    name = "stdout"
+    label_timed_task(name)
+    send_stream(name)
+end
+
+function send_stderr(t::Timer)
```
stevengj: It seems like `send_stdout`, `send_stderr`, and `label_timed_task` can be combined into a single function:
```jl
function send_stdio(t::Timer, name)
    if verbose::Bool && !haskey(task_local_storage(), :IJulia_task)
        task_local_storage(:IJulia_task, "send $name task")
    end
    send_stream(name)
end
```

stevengj: However, I don't think this is the right function to set `task_local_storage` from, as this should normally be called from a function where that has already been set.

So, you might as well eliminate this function entirely and call `send_stream` directly.

src/stdio.jl
JobJob: Yeah, actually I can remove the whole function.

src/stdio.jl
JobJob: Good pickup

src/stdio.jl
JobJob: Indeed, good man. I was toying around with some other ideas with macros there, when I finally went with something simpler, I clearly forgot to de-duplicate.

src/stdio.jl
stevengj: Maybe, but is it the sort of thing that we need permanently in the codebase?

src/stdio.jl
```diff
@@ -52,6 +78,34 @@ function watch_stream(rd::IO, name::AbstractString)
...
+    if buf.size > 0
+        s = takebuf_string(buf)
+        if isvalid(s)
+            send_ipython(publish,
+                     msg_pub(execute_msg, "stream",
+                             @compat Dict("name" => name, "text" => s)))
+        else
+            #if the string is not valid, maybe it's the start of a partially encoded UTF character
+            #so we write it back to the buffer and hope the rest of the character will arrive soon
+            @vprintln("Invalid UTF string written to ", name, ". Bytes: ", convert(Vector{UInt8}, s))
```
stevengj: I don't like having debugging printfs permanently in the code...

Also, this solution is neither efficient nor robust.  It is not efficient because it would be better just to check the end of the string for a partially encoded UTF-8 character, and write only *that* part of the data back to the buffer, sending the rest of the string to Python.   It's not robust because it doesn't handle the case where the use just calls `write(STDOUT, ....raw binary data that is not UTF-8....)`, in which case we should still probably send *something* to IPython.

src/stdio.jl
JobJob: If we want accurate times on the logs, then yes :) I'm personally pro the accurate times, they were useful to me here, and may be useful down the track. However, I'm not going to throw a wobbly delete the whole pull request if you feel differently :) happy to accept whatever you decide. Btw in my other branch I moved these and `vprintln`,`verror_show` to a separate debug.jl file, which may be more palatable?

src/stdio.jl
JobJob: Oh, duh. Just checking the last char is a much more sane way to do it.
PS I put the debug in because I was worried about people wondering where all their output went in case of an invalid char, so will remove it now.

src/stdio.jl
stevengj: In the short run, I'd rather just use the simpler code using `now()`; I think for most purposes we won't need logging with sub-second accuracy.

In the long run, it would be good to submit a PR to Base Julia so that `now()` (or some variant thereof) returns milliseconds.  People shouldn't have to resort to ugly code to get a millisecond timestamp in local time.

src/stdio.jl
```diff
@@ -52,6 +77,36 @@ function watch_stream(rd::IO, name::AbstractString)
...
+        s = takebuf_string(buf)
+        #if the last char is not valid, maybe only the start of a partially encoded UTF character
+        #was written, so we write it back to the buffer in the hope that the rest of the character will arrive soon
+        last_ind = chr2ind(s, length(s))
+        last_chrstr = s[last_ind:sizeof(s)]
+        if !isvalid(last_chrstr)
+            write(buf, last_chrstr)
+            last_ind -= 1
+        end
+        if last_ind > 0
```
stevengj: I'm thinking something more like:
```jl
d = takebuf_array(buf)
n = num_utf8_trailing(d)
if n > 0
    write(buf, d[end-(n-1):end])
    resize!(d, length(d) - n)
end
s = UTF8String(d)
!isvalid(s) && s = base64encode(d) # fallback: base64-encode non-UTF8 binary data
```
where we define the following function to compute the leftover bytes in O(1) time:
```
function num_utf8_trailing(d::Vector{UInt8})
    i = length(d)
    # find last non-continuation byte in d:
    while i >= 1 && ((d[i] & 0xc0) == 0x80)
        i -= 1
    end
    i < 1 && return 0
    c = d[i]
    # compute number of expected UTF-8 bytes starting at i:
    n = c <= 0x7f ? 1 : c < 0xe0 ? 2 : c < 0xf0 ? 3 : 4
    nend = length(d) + 1 - i # num bytes from i to end
    return nend == n ? 0 : nend
end
```

JobJob: Oh I see, you want people do be able to display non-string data - was that possible previously? In any case that's a good idea, and your functions look sexy. Do you want me to just add them and see how they go?

src/stdio.jl
JobJob: Yeah I opened an issue last week https://github.com/JuliaLang/julia/issues/14088 wasn't 100% about the implementation though I tried out changing the c code instead, and it was no faster - do you think I should just submit a PR?

JobJob:
Ok see how you like it, I put your binary handling functions in, and simplified the log timestamp.
One issue I can think of with the binary handling is that one bad character corrupts everything, and it will switch to binary

    write(STDOUT, "this was going to be a great message until... ", 0x99, "so yeah...")

outputs:

    base64 binary data: dGhpcyB3YXMgZ29pbmcgdG8gYmUgYSBncmVhdCBtZXNzYWdlIHVudGlsLi4uIJlzbyB5ZWEuLi4=

With the old version that would throw a `UnicodeError: invalid character index` on the send so I think this still definitely an improvement on that.

src/stdio.jl
stevengj: Julia certainly allows people to do `write(STDOUT, x)` for arbitrary `x`, which will often write binary data.   Therefore have to handle that case *somehow*, even if it is just to print an error message.  Jupyter requires us to send the data as a UTF8 string, so base64 seems as good as anything else.

If you could try out that code and see how it goes, that would begreat.

stevengj:
@JobJob, I think that is sensible; I don't see what we can do with a mix of binary and string data except to treat it as binary.   Certainly it is better than throwing an exception internally.

stevengj:
Does it work correctly with your `write(STDOUT, bathbytes[1:2])` example?  (I wish we had a real test suite; see #327.)

src/stdio.jl
```diff
@@ -52,6 +65,63 @@ function watch_stream(rd::IO, name::AbstractString)
...
+        n = num_utf8_trailing(d)
+        dextra = d[end-(n-1):end]
+        resize!(d, length(d) - n)
+        s = UTF8String(d)
+        if isvalid(s)
+            write(buf, dextra) # assume that the rest of the string will be written later
+            length(d) == 0 && return
+        else
+            # fallback: base64-encode non-UTF8 binary data
+            length(d) == 0 && return
```
stevengj: This test is redundant (it will always be `false`), since `UTF8String(UInt8[])` produces a valid string.

stevengj:
Delete the redundant test mentioned above, and squash the commits, and it looks good to merge!  Thanks for persisting on this!

src/stdio.jl
JobJob: Good. I actually checked that `isvalid("") == true` earlier... too many hours at the screen today :)

JobJob:
Yeah it works with the bathbytes :)
Agree about tests, will look into it.
N.b. I've only tried this on OSX.
Thank you!

JobJob:
n.b. updated commit message.

JobJob:
Nice! Thank you for all your great efforts with this PR but more importantly the package and Julia in general.

Two qs:
 1. Not sure if you missed this comment hidden above, but re `now()` I opened an issue last week JuliaLang/julia#14088 wasn't 100% about the implementation though I tried changing the c code instead, and it was no faster - do you think I should just submit a PR? Ping the relevant authorities again?

 1. Have you had a scan through the issues lately, was wondering if some of them may be old and no longer relevant?

stevengj:
On the second point, I did a quick scan and closed a bunch of obsolete issues; thanks for the reminder.

JobJob:
Nice

PR #404, opened 2016-02-06
0.5 deprecations: super -> supertype and readbytes -> read

schmrlng:
IMPORTANT NOTE: This PR should not be merged until a new version of Compat (currently v0.7.8) is tagged/published.

Just getting a jump on a few upcoming deprecations for Julia v0.5. I've checked that these changes seem to work on both release-0.4 and master; I guess a more robust testing framework is tricky for this sort of package.

### DIFF
--- a/src/stdio.jl
+++ b/src/stdio.jl
@@ -45,7 +45,7 @@ function watch_stream(rd::IO, name::AbstractString)
         while !eof(rd) # blocks until something is available
             nb = nb_available(rd)
             if nb > 0
-                write(buf, readbytes(rd, nb))
+                write(buf, read(rd, nb))
             end
             if buf.size > 0
                 if buf.size >= max_bytes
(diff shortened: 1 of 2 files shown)


stevengj:
The PR should also update the Compat version number in the `REQUIRE` file, when a new version is available.

schmrlng:
Okay I think this should be ready to go.

src/stdio.jl
```diff
@@ -45,7 +45,7 @@ function watch_stream(rd::IO, name::AbstractString)
         while !eof(rd) # blocks until something is available
             nb = nb_available(rd)
             if nb > 0
-                write(buf, readbytes(rd, nb))
+                write(buf, @compat(read(rd, nb)))
```
stevengj: You only need the `@compat` macro for syntax changes.   For just function-name changes, it should work fine without `@compat`.

stevengj:
Thanks!

PR #413, opened 2016-04-07
Fix depwarn of readall in deps/build.jl

yuyichao:


yuyichao:
Appveyor seems to have some issue with `Conda.jl` and then hit https://github.com/JuliaLang/julia/issues/15789 . Restart to see if the issue is gone (in any case it should be unrelated to this PR)...

PR #422, opened 2016-05-12
Fix depwarn on 0.5

yuyichao:


yuyichao:
Yes, and fixed in https://github.com/JuliaLang/IJulia.jl/commit/fcbcef24d2b022464092c1d0f99555edc27973a0

PR #431, opened 2016-06-17
Add application/javascript as a MIME type

bluesmoon:
IPython/Jupyter already know how to handle `application/javascript`, so we should provide that as a usable MIME type


stevengj:
Sounds good.  What does Jupyter do with application/javascript code?

bluesmoon:
it does an `eval` on the code.  This way the kernel app can send `application/javascript` that is executed immediately without creating a new script node in the DOM, and without creating a container element which takes up visual space.

PR #432, opened 2016-06-23
comm_info_request

shashi:
This is required to fix Interact on the new ipywidgets.

see https://github.com/ipython/ipywidgets/pull/175


shashi:
It looks like execution time exceeded limit on AppVeyor...

PR #441, opened 2016-07-17
Fix error on 0.5-dev due to deleted isvalid method

yuyichao:


stevengj:
It seems a little silly that isvalid(x::T) does not default to isvalid(T, x)

stevengj:
(Assuming that we insist on passing the type explicitly at all.)

Issue #443, opened 2016-08-06
Running any cell in IJulia notebook gives bytestring(zmsg::Message) deprecated warning


c123w:
If I make a cell with (for example) a=1, and then run it, I get an output of 1 as expected, but I also get the following error message:

```
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
WARNING: bytestring(zmsg::Message) is deprecated, use unsafe_string(zmsg::Message) instead.
 in depwarn at ./deprecated.jl:73
while loading /home/johnsmith/.julia/v0.4/IJulia/src/kernel.jl, in expression starting on line 35
```
Why am I getting this error? Is it actually anything to worry about? (and if not, how can I suppress this warning?)

I'm  running a fresh install of Julia/IJulia (0.4.6) on Ubuntu 16.

jmbromley:
[...]

PR #444, opened 2016-08-08
use unsafe_string instead of bytestring

wookay:
It was a closed issue but to fix the potential problem.
https://github.com/JuliaLang/IJulia.jl/issues/443

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.4
 Nettle 0.2
 JSON 0.5
-ZMQ 0.3.1
+ZMQ 0.3.3
 Compat 0.7.20
 Conda 0.1.5
--- a/src/msg.jl
+++ b/src/msg.jl
@@ -71,21 +71,20 @@ function recv_ipython(socket)
     try
         msg = recv(socket)
         idents = String[]
-        # unsafe_string for ZMQ.jl has patched
-        s = bytestring(msg)
+        s = unsafe_string(msg)
         @vprintln("got msg part $s")
         while s != "<IDS|MSG>"
             push!(idents, s)
             msg = recv(socket)
-            s = bytestring(msg)
+            s = unsafe_string(msg)
             @vprintln("got msg part $s")
         end
-        signature = bytestring(recv(socket))
+        signature = unsafe_string(recv(socket))
         request = Dict{String,Any}()
-        header = bytestring(recv(socket))
-        parent_header = bytestring(recv(socket))
-        metadata = bytestring(recv(socket))
-        content = bytestring(recv(socket))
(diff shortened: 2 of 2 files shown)


wookay:
reviewing with https://github.com/JuliaLang/ZMQ.jl/pull/109

wookay:
it's planning for ZMQ.jl versions https://github.com/JuliaLang/ZMQ.jl/issues/110

REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.4
 Nettle 0.2
 JSON 0.5
-ZMQ 0.3.1
+ZMQ 0.3.4
```
tkelman: is this fixing a deprecation warning or an error? if it's just fixing a deprecation warning, it's too tight

wookay: this should be ZMQ 0.3.4.
it's been changed with `unsafe_string(::Message)` since ZMQ 0.3.3,
but ZMQ 0.3.3 has wrong(conditional deprecation).

tkelman: It works, it just has a lot of warnings. I suspect versions earlier than 0.3.1 would work too without any errors, wouldn't they?

wookay: I mean, for Julia 0.4, ZMQ 0.3.1 was defined for the type `Message` with
```
bytestring(zmsg::Message) = bytestring(pointer(zmsg), length(zmsg))
```
and with this PR, calling `unsafe_string(::Message)` doesn't work on Julia 0.4.

when running jupyter-notebook, there's hanging with timeout like this
```
[W 21:35:58.097 NotebookApp] Timeout waiting for kernel_info reply from e3f8da4c-ce1c-4604-965e-200479992642
```

thanks.

tkelman: Ah, that makes sense, there was no `unsafe_string(::Message)` method before ZMQ.jl 0.3.3? So ZMQ.jl 0.3.3 is the actual minimum.

wookay: that's also bad, because there's the critical issue for on ZMQ 0.3.3.
https://github.com/JuliaLang/ZMQ.jl/issues/108#issuecomment-238099002
incidentally causing lot's of deprecated warnings for every input on IJulia.
I prefer to use ZMQ 0.3.4.

tkelman: The requirements of ZMQ 0.3.3 and 0.3.4 are the same, so any time both are possible, Pkg will use the larger of them.

stevengj:
Bump.  Needs a rebase and updated REQUIRE to ZMQ 0.3.3

wookay:
ok. I'm going to rebase to ZMQ 0.3.3

stevengj:
Thanks.

PR #447, opened 2016-08-14
Switch to MbedTLS

malmaud:
Since Julia ships with MbedTLS binaries now, I'm trying to switch over packages from Nettle.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.4
-Nettle 0.2
+MbedTLS 0.3.0
 JSON 0.5
 ZMQ 0.3.3
 Compat 0.9.5
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
-julia 0.4
-MbedTLS 0.3.0
+julia 0.5
+MbedTLS 0.4.3
 JSON 0.5
 ZMQ 0.3.3
 Compat 0.9.5
--- a/src/init.jl
+++ b/src/init.jl
@@ -64,7 +64,7 @@ function init(args)
         if signature_scheme[1] != "hmac" || length(signature_scheme) != 2
             error("unrecognized signature_scheme $signature_scheme")
         end
-        push!(hmacstate, HMACState(signature_scheme[2], profile["key"]))
+        hmacstate[] = MbedTLS.MD(MbedTLS.MD_SHA256, profile["key"])
     end
 
     ctx[] = Context()
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.4
   - 0.5
   - nightly
 notifications:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,5 @@
 environment:
   matrix:
-  - JULIAVERSION: "julialang/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
-  - JULIAVERSION: "julialang/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIAVERSION: "julialang/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIAVERSION: "julialang/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
   - JULIAVERSION: "julianightlies/bin/winnt/x86/julia-latest-win32.exe"
--- a/src/hmac.jl
+++ b/src/hmac.jl
@@ -1,13 +1,15 @@
-using Nettle
-const hmacstate = HMACState[]
+using MbedTLS
+const hmacstate = Ref{MbedTLS.MD{true}}()
+
 function hmac(s1,s2,s3,s4)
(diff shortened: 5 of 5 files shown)


malmaud:
As far as I can tell, the AppVeyor failure doesn't seem related to this PR.

stevengj:
Julia 0.4 has this?

malmaud:
mbedtls doesn't ship with .4, but MbedTLS will still build(/binaries will be fetched) fine on .4.

stevengj:
Yeah, looks like a separate AppVeyor problem (with Conda).

src/IJulia.jl
```diff
@@ -15,17 +15,19 @@ const capture_stderr = true
...
 using JSON
 
-using Nettle
-const hmacstate = HMACState[]
+using MbedTLS
+const hmacstate = Ref{MbedTLS.MD{true}}()
+
 function hmac(s1,s2,s3,s4)
-    if isempty(hmacstate)
+    if !isdefined(hmacstate, :x)
```
stevengj: Julia should really have an `isdefined(x::Ref) = isdefined(x, :x)` method...

src/IJulia.jl
```diff
@@ -15,17 +15,19 @@ const capture_stderr = true
...
-        update!(hmacstate[1], s2)
-        update!(hmacstate[1], s3)
-        update!(hmacstate[1], s4)
-        return hexdigest!(hmacstate[1])
+        MbedTLS.reset!(hmacstate[])
+        for s in (s1, s2, s3, s4)
+            write(hmacstate[], s)
+        end
+        # Take the digest (returned as a byte array) and convert it to hex string representation
+        return join([hex(_, 2) for _ in MbedTLS.finish!(hmacstate[])])
```
stevengj: maybe
```
digest = MbedTLS.finish!(hmacstate[])
hexdigest = Array(UInt8, length(digest)*2)
for i = 1:length(digest)
    b = digest[i]
    d = b >> 4
    hexdigest[2i-1] = UInt8('0')+d+39*(d>9)
    d = b & 0xf
    hexdigest[2i] = UInt8('0')+d+39*(d>9)
end
return Compat.ASCIIString(hexdigest)
```
to avoid all of the temporary allocations.

stevengj: (Though maybe it's not worth it to be fancy here.)

src/IJulia.jl
malmaud: It really should.

src/IJulia.jl
malmaud: My vote is not to be fancy here given this isn't on a performance-critical path, but whatever you think is best.

stevengj: Since this is called for every message, I'd rather do the fancy thing.

stevengj:
Bump.

tkelman:
what is wrong with nettle exactly?

malmaud:
It's just one more binary dependency that can cause trouble. With mbedtls
bundled with Julia now, it seems the simplest.

tkelman:
except that mbedtls.jl doesn't use the bundled library yet, so you're trading one download for another.

malmaud:
It's just a matter of getting that PR in

tkelman:
by "in" you mean "working" right? ref https://github.com/JuliaWeb/MbedTLS.jl/pull/66

malmaud:
I believe MbedTLS now indeed correctly uses the pre-shipped binaries, so time to revisit this.

tkelman:
only on Julia 0.5

malmaud:
Still, MbedTLS will just build/download as a normal external dependency on .4 as Nettle does now. Or were you thinking on Julia <.5 IJulia would still use Nettle? For that, I guess we'd to tag a minor version of IJulia that bumps the Julia requirement to .5.

tkelman:
That's probably worth doing when making this change, rather than trying to worry about one more permutation in which we need to keep things working. If Nettle works fine on 0.4, probably better to leave it be.

REQUIRE
```diff
@@ -1,5 +1,5 @@
-julia 0.4
```
tkelman: also in travis, if stevengj agrees about now (well, after #502 gets merged and tagged first) being an okay time to drop 0.4

stevengj: Yes, I think dropping 0.4 should be fine.

malmaud:
@stevengj I think this is good to go now

src/init.jl
```diff
@@ -64,7 +64,7 @@ function init(args)
         if signature_scheme[1] != "hmac" || length(signature_scheme) != 2
             error("unrecognized signature_scheme $signature_scheme")
         end
-        push!(hmacstate, HMACState(signature_scheme[2], profile["key"]))
+        hmacstate[] = MbedTLS.MD(MbedTLS.MD_SHA256, profile["key"])
```
stevengj: I just noticed that this assumes `signature_scheme[2] == "sha256"`

malmaud:
I'll take a look today

src/init.jl
stevengj: Fixed in 7c8fe35fe2e61ab231defdb1fbe304ab275f6db7

PR #459, opened 2016-09-07
Explain that detached option doesn't block current process also.

Ismael-VC:
Discussion reference in julia chat room:

* https://gitter.im/JuliaLang/julia?at=57d034e877c1b70d7ff1cccb


yuyichao:
Why not `notebook(detached=true)`

Ismael-VC:
Yes I will change it to show example output and state it also does not block the process, thanks again!

PR #461, opened 2016-09-11
Fix use of `invoke`

yuyichao:
* Use tuple type instead of tuples (Ref https://github.com/JuliaLang/julia/pull/18444 and it's easier to infer in general)
* Use `eval` instead of `invoke` to see newly defined methods. The old method won't work after https://github.com/JuliaLang/julia/pull/18444 or after JuliaLang/julia#265 is fixed.


stevengj:
Thanks.  (Travis failure seems to be an unrelated Conda download timeout.)

PR #463, opened 2016-09-15
truncate excess output per execution request

### LINKED ISSUE #455: Limit show/display output?
Would it be possible to limit the maximum lines from `println` and `show`? I find that I often stall out `IJulia` when accidentally printing a too large array and then it is annoying and painful to clear the output without the browser going down. Say 1000 lines?

tlnagy:
fixes https://github.com/timholy/Images.jl/issues/558, fixes #455

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,3 +250,13 @@ your `.julia` package directory) to change the line `verbose = false`
 at the top to `verbose = true` and `const capture_stderr = true` to
 `const capture_stderr = false`.  Then restart the kernel or open a new
 notebook and look for the error message when IJulia dies.
+
+### Preventing truncation of output
+
+The new default behavior of IJulia is to truncate stdout (via `show` or `println`)
+after 512kb. This to prevent browsers from getting bogged down when displaying the
+results. This limit can be increased to a custom value, like 1MB, as follows
+
+```
+IJulia.set_max_excessive_output(1 << 20)
+```
--- a/README.md
+++ b/README.md
@@ -258,5 +258,5 @@ after 512kb. This to prevent browsers from getting bogged down when displaying t
 results. This limit can be increased to a custom value, like 1MB, as follows
 
 ```
-IJulia.set_max_excessive_output(1 << 20)
+IJulia.set_max_stdio(1 << 20)
 ```
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -286,15 +286,14 @@ end
 
 
 """
-    set_max_excessive_output(max_output::Int)
+    set_max_stdio(max_output::Integer)
 
-Sets the maximum number of bytes, `max_output`, that can be written to stdout
-before getting truncated. A large value here allows a lot of output to be
+Sets the maximum number of bytes, `max_output`, that can be written to stdout and
+stderr before getting truncated. A large value here allows a lot of output to be
 displayed in the notebook, potentially bogging down the browser.
 """
-function set_max_excessive_output(max_output::Int)
+function set_max_stdio(max_output::Integer)
     max_output_per_request[] = max_output
-    println("Stdout will now be truncated after $max_output bytes")
 end
 
 
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -284,6 +284,20 @@ function clear_output(wait=false)
                                     Dict("wait" => wait)))
 end
 
+
+"""
+    set_max_excessive_output(max_output::Int)
+
+Sets the maximum number of bytes, `max_output`, that can be written to stdout
+before getting truncated. A large value here allows a lot of output to be
+displayed in the notebook, potentially bogging down the browser.
+"""
+function set_max_excessive_output(max_output::Int)
+    max_output_per_request[] = max_output
+    println("Stdout will now be truncated after $max_output bytes")
+end
+
+
 #######################################################################
 
 include("init.jl")
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -111,6 +111,9 @@ end
 
 # global variable so that display can be done in the correct Msg context
 execute_msg = Msg(["julia"], Dict("username"=>"julia", "session"=>"????"), Dict())
+# global variable tracking the number of bytes written in the current execution
+# request
+const stdout_bytes = Ref(0)
 
 function helpcode(code::AbstractString)
     code_ = strip(code)
@@ -132,6 +135,7 @@ function execute_request(socket, msg)
     @vprintln("EXECUTING ", code)
     global execute_msg = msg
     global n, In, Out, ans
+    stdout_bytes[] = 0
     silent = msg.content["silent"]
     store_history = get(msg.content, "store_history", !silent)
     empty!(execute_payloads)
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -113,7 +113,7 @@ end
 execute_msg = Msg(["julia"], Dict("username"=>"julia", "session"=>"????"), Dict())
 # global variable tracking the number of bytes written in the current execution
 # request
-const stdout_bytes = Ref(0)
+const stdio_bytes = Ref(0)
 
 function helpcode(code::AbstractString)
     code_ = strip(code)
@@ -135,7 +135,7 @@ function execute_request(socket, msg)
     @vprintln("EXECUTING ", code)
     global execute_msg = msg
     global n, In, Out, ans
-    stdout_bytes[] = 0
+    stdio_bytes[] = 0
     silent = msg.content["silent"]
     store_history = get(msg.content, "store_history", !silent)
     empty!(execute_payloads)
--- a/src/stdio.jl
+++ b/src/stdio.jl
@@ -31,12 +31,18 @@ end
 #name=>iobuffer for each stream ("stdout","stderr") so they can be sent in flush
 const bufs = Dict{String,IOBuffer}()
 const stream_interval = 0.1
+# maximum number of bytes in libuv/os buffer before emptying
 const max_bytes = 10*1024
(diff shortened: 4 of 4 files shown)


tlnagy:
Not sure why Appveyor is failing, looks unrelated to this PR. @stevengj what do you think?

src/stdio.jl
```diff
@@ -32,20 +32,41 @@ end
 const bufs = Dict{String,IOBuffer}()
 const stream_interval = 0.1
 const max_bytes = 10*1024
+# max output per code cell is 512 kb
+const max_output_per_request = 1 << 19
```
stevengj: This shouldn't be a constant, it should be a `Ref(1 << 19)` so that we can provide an API to change it in a running notebook.

src/stdio.jl
```diff
@@ -32,20 +32,41 @@ end
...
 function watch_stream(rd::IO, name::AbstractString)
     task_local_storage(:IJulia_task, "read $name task")
     try
         buf = IOBuffer()
         bufs[name] = buf
+        counter = 0
+        curr_msg = ""
         while !eof(rd) # blocks until something is available
             nb = nb_available(rd)
+            parent_id_exists = "msg_id" in keys(execute_msg.header)
```
stevengj: I would just put the counter as a global `Ref(0)` and reset it to zero in the execute_request handler. Then all this logic about checking message ids can be removed, and we have a single counter for all streams.

src/stdio.jl
```diff
@@ -32,20 +32,41 @@ end
...
+            if parent_id_exists && !continuation
+                curr_msg = execute_msg.header["msg_id"]
+                counter = 0
+            end
             if nb > 0
-                write(buf, read(rd, nb))
+                counter += nb
+                # if this stream as surpassed the maximum output limit then ignore future bytes
+                if parent_id_exists && continuation && counter > max_output_per_request
+                    (counter - nb <= max_output_per_request) && println(orig_STDOUT[], "Excessive output truncated.")
```
stevengj: it is weird to have the `println` as part of the conditional, especially since `println` returns `nothing` ... won't this give an error?  `if true && println("foo"); end` gives an error `TypeError: non-boolean (Void) used in boolean context`

I don't think we want to print a message on every read; I think we just want one warning per cell, and I think we want to print it to the notebook.   If you switch to the global counter as I suggested above, we can just do:
```
if stdio_counter[] > stdio_max[]
   counter[] = 0
   warn("Output > ", stdio_max[], " bytes was truncated.")
end
```
in when the execute completes.

src/stdio.jl
tlnagy: Your global counter is likely a better solution.

src/stdio.jl
tlnagy: That's pretty standard Julia syntax, Stefan even recommends it here: https://stackoverflow.com/questions/22959499/julia-can-conditional-statements-evaluate-code-on-the-same-line and I see it all over the place.

The problem is that the best place to evaluate the number of bytes written is in the primary event loop. Your solution wouldn't work because it does not actually truncate future bytes which are being streamed to us. Your suggestion of resetting the counter `execute_handler` would likely work. My opposing conditionals ensured that we only print a warning message right when we overstep `max_output_per_request`, but keeps read from the libuv/os buffer so that the bytes actually get truncated, i.e. read and discarded.

I don't write it to the notebook because I'm truncating stdout and the truncation includes my warning message too. Maybe there is a more clever way of handling this.

stevengj: @tlnagy, nevermind, I was reading the `println` as part of the same conditional as the `if` on the previous line.  I agree that `condition && println()` is a standard idiom.  I misread it as doing `if condition && println()`, however, which would have been wrong.

stevengj: Right, you only want to reset the execution counter at the *beginning* of the execution loop, not at the end.

However, I would still like a warning to be printed in the notebook itself, not in the terminal where the `jupyter` server is running.

tlnagy: One way to get that too work is to make sure that we are only truncating in when `name = "stdout"` while allowing `stderr` to pass through. Then we could dump the error to `stderr` and the user should be able to see it.

src/stdio.jl
tlnagy: Why use `Ref` here?

yuyichao: Because

> so that we can provide an API to change it in a running notebook.

tlnagy:
I made all the suggested changes, added a function for changing the limit, and added docs to the README.

src/stdio.jl
tlnagy: Either way we need a nice function for changing it so it shouldn't matter too much whether it's a global or a `Ref`. I ended up going with the latter.

yuyichao: You can't change it if it's a const.

stevengj: Julia knows the type of a `const foo = Ref(bar)`; this is the advantage of `Ref` over a non-`const` global.

src/stdio.jl
stevengj: No, I think we should throttle both stdio streams.   Either one could have runaway output.

src/stdio.jl
tlnagy: Ah. That makes sense. I ended up using `Ref` anyway, but good to know.

src/stdio.jl
tlnagy: I'm not sure how to throttle stderr while allowing and a truncation warnings through. Currently, I just call `warn` once stdout has passed the max limit and then just read and discard future output. If I don't have the check for `stdout` then the `warn` message also gets truncated and you can't see it.

src/stdio.jl
```diff
@@ -45,7 +51,20 @@ function watch_stream(rd::IO, name::AbstractString)
         while !eof(rd) # blocks until something is available
             nb = nb_available(rd)
             if nb > 0
-                write(buf, read(rd, nb))
+                if name == "stdout"
+                    stdout_bytes[] += nb
+                    # if this stream as surpassed the maximum output limit then ignore future bytes
```
JobJob: typo: if this stream _**has**_ not as

README.md
```diff
@@ -250,3 +250,13 @@ your `.julia` package directory) to change the line `verbose = false`
 at the top to `verbose = true` and `const capture_stderr = true` to
 `const capture_stderr = false`.  Then restart the kernel or open a new
 notebook and look for the error message when IJulia dies.
+
+### Preventing truncation of output
+
+The new default behavior of IJulia is to truncate stdout (via `show` or `println`)
```
stevengj: stdout and stderr

README.md
```diff
@@ -250,3 +250,13 @@ your `.julia` package directory) to change the line `verbose = false`
...
+after 512kb. This to prevent browsers from getting bogged down when displaying the
+results. This limit can be increased to a custom value, like 1MB, as follows
+
+```
+IJulia.set_max_excessive_output(1 << 20)
```
stevengj: This name is excessive.  😉   Just `set_max_stdio` should be enough.

src/IJulia.jl
```diff
@@ -284,6 +284,20 @@ function clear_output(wait=false)
...
 
+
+"""
+    set_max_excessive_output(max_output::Int)
+
+Sets the maximum number of bytes, `max_output`, that can be written to stdout
+before getting truncated. A large value here allows a lot of output to be
+displayed in the notebook, potentially bogging down the browser.
+"""
+function set_max_excessive_output(max_output::Int)
```
stevengj: `::Integer`

src/execute_request.jl
```diff
@@ -111,6 +111,9 @@ end
 
 # global variable so that display can be done in the correct Msg context
 execute_msg = Msg(["julia"], Dict("username"=>"julia", "session"=>"????"), Dict())
+# global variable tracking the number of bytes written in the current execution
+# request
+const stdout_bytes = Ref(0)
```
stevengj: `stdio_bytes`, since we should track both stdout and stderr.

src/stdio.jl
```diff
@@ -45,7 +51,20 @@ function watch_stream(rd::IO, name::AbstractString)
         while !eof(rd) # blocks until something is available
             nb = nb_available(rd)
             if nb > 0
-                write(buf, read(rd, nb))
+                if name == "stdout"
```
stevengj: Do this regardless of stream name, i.e. limit both stdout and stderr

src/stdio.jl
```diff
@@ -45,7 +51,20 @@ function watch_stream(rd::IO, name::AbstractString)
...
+                    stdout_bytes[] += nb
+                    # if this stream as surpassed the maximum output limit then ignore future bytes
+                    if stdout_bytes[] > max_output_per_request[]
+                        if stdout_bytes[] - nb <= max_output_per_request[]
+                            warn("Excessive output truncated after $(stdout_bytes[]) bytes.")
```
stevengj: See @JobJob's suggestion of just calling `send` directly

src/stdio.jl
tlnagy: That doesn't seem to work in my hands. The output doesn't get displayed and the kernel hangs.

JobJob: Sorry, I was looking at an old version of the code, does this work:

    send_ipython(publish[],
         msg_pub(execute_msg, "stream",
                Dict("name" => "stderr", "text" => "Excessive output truncated after $(stdout_bytes[]) bytes.")))

Am just copying from here: https://github.com/JuliaLang/IJulia.jl/blob/master/src/stdio.jl#L100-L102

Btw, I'm very happy you're implementing this feature, it often causes me problems.

tlnagy:
Thanks @JobJob and @stevengj for the review, I've made the requested changes.

PR #472, opened 2016-09-21
peel off one LoadError from thrown error

### LINKED ISSUE #471: Every error in IJulia is a LoadError
image

I guess because the text gets `include`d? Perhaps the error should be unpacked before getting thrown.

KristofferC:
fixes #471 

image

image


stevengj:
Does the REPL do something similar?

KristofferC:
I think the REPL evals the AST instead of `include_string`.

PR #476, opened 2016-10-04
stop showing deprecations in notebooks

KristofferC:
Does anyone actually think showing deprecations in notebooks is useful? Development on packages is done in editors. The only thing depwarns do is serve as a source of annoyment when you code yourself and embarrassment when a package emits 500 depwarnings when you have a presentation using IJulia.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -62,7 +62,7 @@ juliakspec = abspath(spec_name)
 binary_name = is_windows() ? "julia.exe" : "julia"
 kernelcmd_array = String[joinpath(JULIA_HOME,("$binary_name")), "-i"]
 ijulia_dir = get(ENV, "IJULIA_DIR", Pkg.dir("IJulia")) # support non-Pkg IJulia installs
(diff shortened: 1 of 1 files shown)


stevengj:
Fair enough.

stevengj:
On the other hand, I would prefer to leave depwarns on for code that is in the notebook itself; this is important to teach users about syntax changes.

Is there any way to disable depwarns for modules but leave it on for code in `Main`?

KristofferC:
Not that I know. The `@deprecate` macro just defines a new function that emits the dep warning and calls the new function so it is pretty context insensitive. I do see your point about notifying users of packages of deprecated syntax they are using.

What would you think about a PR that gives `notebook` an extra optional argument of type `Dict{String, String}` so you can give extra keywords yourself. Like `notebook(;args = Dict("depwarn" => "no"))`

stevengj:
@KristofferC, I don't know that there is a good Jupyter API for that, short of overwriting the kernel file every time you call `notebook`.

KristofferC:
Ah, I would have assumed there was a way but you are right, you can only specify them in the kernel file. Bummer.

KristofferC:
Guess we have to figure out some other way then.

PR #490, opened 2016-12-12
Add a flush_all to the end of the eventloop

JobJob:
Some Interact/Reactive stuff is async now, and I think putting that flush can help slightly with lost output, particularly on errors, for example during user widget interaction.

### DIFF
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -26,6 +26,7 @@ function eventloop(socket)
                                  msg_pub(execute_msg, "error", content))
                 end
             finally
+                flush_all()
                 send_status("idle", msg.header)
             end
         end


stevengj:
Running this after each message is processed seems kind of random to me.

JobJob:
Yeah fair enough. I figured it might help others using their own Comms like PlotlyJS, but I can try narrow down a case where it helps, and be a little more clear about why. The thing is Jupyter stops showing stream messages shortly after the execution_state: idle is message is received.

stevengj:
@minrk, is there any way to tell Jupyter that we might have I/O from background tasks and it shouldn't ignore stream messages in the idle state?  (Why should it do this in the first place?)

JobJob:
I looked into this a little further, in the normal case, stream messages arriving after the `"execution_state":"idle"` are displayed under the cell last executed. 

However, in the case where we are returning or displaying a Reactive Signal, some subscribe_mime messages arrive, and the kernel sends some intervening "busy", "idle" messages. After that, stream messages sent aren't displayed.

@minrk could the issue be something like: the front-end is trying to display the stream message under a cell associated with the last message, but in this case the last messages are the subscribe-mimes and so there is no associated cell?

N.b. The stream messages are arriving at the front-end (as reported in the browser js console network tab).

---
Here is a minimal example:
```
using Interact
p = Signal(0.0)
push!(p, "1") #error because "1" is a String and p only accepts Float64
p
```
The error message gets truncated. If you change the last line to `display(p);` you get the same outcome. However, if you change it to `p;` so the Signal `p` isn't displayed, you get the full error message.

minrk:
The `status: idle` message is a soft indicator that output is done for a given request. The easiest way to guarantee that output will still be handled is to prevent that idle message from being sent, if you know that async output will still come that should be associated with the given request (e.g. IPython does `sys.stdout.flush()` prior to sending status: idle).

The reason messages are lost is that the notebook frontend does some garbage collection, to remove listeners for events once status: idle is received. Async output is handled in that the callbacks for the most recent request are always kept around, so async output associated with the last cell will show up, but arbitrarily delayed outputs still targeted for other cells will have lost their destination. JupyterLab takes (will take?) a better approach, in that the handlers for every cell on the page will remain active as long as the cell is there, so the active handlers are bounded by the number of cells on the page.

JobJob:
Thanks for that @minrk. Glad to hear JupyterLab will improve things.

In this particular case, after the cell executes and the kernel sends the `status:idle`, we send a message from the front-end to the kernel on a custom `"comm_msg"` channel. This new message causes Jupyter to overwrite all the `last_msg_callbacks` and since the callbacks associated with the previous cell's (parent) `msg_id` are removed when the `status:idle` is received, there are no appropriate callbacks to write the stream messages to an output area and they're effectively dropped.

Closing this PR, it doesn't actually fix the problem. I have a possible work-around in #491

PR #491, opened 2016-12-15
Add mechanism to delay sending the idle message.

JobJob:
Jupyter removes IO handling callbacks associated with a cell when the `"status":"idle"` message with the (parent header's) `msg_id` associated with the cell is received. New (asynchronous) output stream messages received after the `"status":"idle"` will still be routed to the last active cell under normal circumstances. However, if a new message is sent by the front-end, the `last_msg_callbacks` that would have put the output on an appropriate cell are overwritten/cleared and subsequent stream messages are silently dropped.

This change will allow other packages to tell IJulia to delay sending the `status:idle`. In particular, it will help when Interact.jl displays a signal but also has an error during an asynchronous `push!` to that or another signal. In that case Interact's javascript sends a comm_msg which clears the  `last_msg_callbacks` as described above, and the error message/stacktrace is usually truncated. With this change, Interact can tell IJulia to delay sending the idle by, e.g., 0.5 secs which fixes the problem.

### DIFF
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -26,7 +26,11 @@ function eventloop(socket)
                                  msg_pub(execute_msg, "error", content))
                 end
             finally
-                send_status("idle", msg.header)
+                @async begin
+                    sleep(idle_delay[])
+                    flush_all()
+                    send_status("idle", msg.header)
+                end
             end
         end
     catch e
@@ -55,4 +59,9 @@ function waitloop()
             end
         end
     end
+end
+
+const idle_delay = Ref(0.5)
+function set_idle_delay(delay_secs::Float64=0.5)
+    idle_delay[] = delay_secs
 end


JobJob:
The change to make in Interact to test this is to add this:
```
IJulia.set_delay_idle(0.5)
```
to this line: https://github.com/JuliaGizmos/Interact.jl/blob/master/src/IJulia/setup.jl#L82

The test case is:
```
using Interact
p = Signal(0.0)
push!(p, "1") #error because "1" is a String and p only accepts Float64
p
```
Without the delay idle, the error message/stacktrace is usually truncated

JobJob:
Btw, making the change that it always asynchronously delays 0.5 or even 1.0 second before sending the idle is not the worst solution either and it's more likely to help in future unforseen cases.

We tell Jupyter to route the IO to a particular cell with the `execute_msg` header, so changing it to always asynchronously delay sending the idle won't actually change where the output will be displayed, nor cause any other major problems as far as I can tell.

stevengj:
Let's just always delay 0.5 seconds.

JobJob:
Good. Updated.

stevengj:
This won't impose a delay between cells if you are evaluating multiple cells in a row, will it?

JobJob:
No

JobJob:
Since it's async

JobJob:
And it still will send commands to the kernel even if previous cells are idle. You can try it out with `IJulia.set_idle_delay(5.0)` and then run multiple cells.

JobJob:
👍

Issue #497, opened 2017-01-11
IJulia pendant for IPythons %hist / %history


crstnbr:
I'm using IJulia within qtconsole via

` jupyter qtconsole --kernel julia-0.5`

A usual workflow for me is playing around in this interactive qtconsole and then copy the working code into some script file. When using IPython (instead of IJulia) I can get a simple linear ordered history, which is very nicely copyable, by running `%hist` or `%history`.

I think I figured out that in Julia the best I can do is use IJulia.In dictionary (got that from [magics.jl](https://github.com/JuliaLang/IJulia.jl/blob/bbfd419dbe896b3b5c3b29557c951c47df489338/src/magics.jl#L123), by the way, the relevant help only shows up when executing `%history` and not for `%hist`). However, to get a simple linear ordered history from it I have to fiddle around to get rid of the mapping information ("number"=>) the quotation marks and line breaks (\n) etc.

I would like to have a simple plain text ordered history of the last X executed commands/cells. To avoid keeping another history, it would be enough to provide a function that maps IJulia.In to such a simple history.

stevengj:
`history(n) = print(join(map(n -> In[n], sort(collect(keys(In)))[end-n:end]), "\n\n"))` prints the last n history items, for example.  Is this what you had in mind?

crstnbr:
Yes, thanks! That is more elegant than my personal fiddling solution.
To be precise, I find those two functions very convenient:
```
history(n) = print(join(map(n -> In[n], sort(collect(keys(In)))[end-n:end]), ""))
history() = print(join(map(n -> In[n], sort(collect(keys(In)))[1:end]), ""))
```

Of course, I can define them myself but why not integrate this function and suggest it when someone tries to call the above IPython magic commands? You guys know better then me what to integrate and what not but, if I have a vote, I would vote for integration :)

I would love to hear your (general) comments on this (Should I prepare a pull request? Does it have any chance?). Otherwise this issue can be closed.

stevengj:
It seems reasonable to me to include this, and PR would be welcome.

I would tend to suggest calling it `printhistory`, with an optional `io` argument to output to a different stream:
```jl
"""
documentation
"""
function printhistory(io::IO, n=length(In))
    n = min(n, length(In))
    n < 0 && throw(ArgumentError("a non-negative argument is required"))
    print(join(map(n -> In[n], sort(collect(keys(In)))[end-n+1:end]), "\n\n"))
end
[...]

PR #500, opened 2017-01-16
print_history() function

crstnbr:
As discussed in #497, this adds a print_history function to IJulia that prints the input history similar to IPythons %history magic. Arguments can be an ::IO target and an integer range, array, or tuple to specify the cell number to output. The default print_history() prints all input history.

### DIFF
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -212,7 +212,7 @@ clear_history
 
 #######################################################################
 # methods to print history or any subset thereof
-function print_history(io::IO, indices)
+function history(io::IO, indices)
 	# since a range could be huge, intersect it with 1:n first
     for n in intersect(indices, 1:IJulia.n)
       if haskey(In, n)
@@ -220,16 +220,16 @@ function print_history(io::IO, indices)
       end
     end
 end
-print_history(x) = print_history(Base.STDOUT, x)
-print_history() = print_history(1:n)
+history(x) = history(Base.STDOUT, x)
+history() = history(1:n)
 
 # print the last x history entries when argument x is integer
-print_history(io::IO, x::Integer) =
-    invoke(print_history, Tuple{IO, Any}, io, n-x:n)
+history(io::IO, x::Integer) =
+    invoke(history, Tuple{IO, Any}, io, n-x:n)
 """
-    print_history([io], [indices])
+    history([io], [indices])
 
-The `print_history()` function prints all of the input history stored in
+The `history()` function prints all of the input history stored in
 the running IJulia notebook in a format convenient for copying.
 
 The optional `indices` argument is a collection of indices indicating
@@ -238,7 +238,7 @@ a subset of cell inputs to print.
 The optional `io` argument is for specifying an output stream. The default
 is Base.STDOUT.
 """
-print_history
+history
 
 #######################################################################
 # Similar to the ipython kernel, we provide a mechanism by
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -212,20 +212,18 @@ clear_history
 
 #######################################################################
 # methods to print history or any subset thereof
-function history(io::IO, indices)
-	# since a range could be huge, intersect it with 1:n first
+function history(io::IO, indices::AbstractVector{Int})
     for n in intersect(indices, 1:IJulia.n)
       if haskey(In, n)
         print(In[n])
       end
     end
 end
-history(x) = history(Base.STDOUT, x)
-history() = history(1:n)
 
-# print the last x history entries when argument x is integer
-history(io::IO, x::Integer) =
-    invoke(history, Tuple{IO, Any}, io, n-x:n)
+history(io::IO, x::Union{Integer,AbstractVector{Int}}...) = history(io, vcat(x...))
+history(x...) = history(STDOUT, x...)
+history(io::IO, x...) = throw(MethodError(history, (x...)))
+history() = history(1:n)
 """
     history([io], [indices])
 
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -225,13 +225,13 @@ history(x...) = history(STDOUT, x...)
 history(io::IO, x...) = throw(MethodError(history, (x...)))
 history() = history(1:n)
 """
-    history([io], [indices])
+    history([io], [indices...])
 
 The `history()` function prints all of the input history stored in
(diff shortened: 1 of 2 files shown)


src/IJulia.jl
```diff
@@ -211,6 +211,50 @@ a subset of cell inputs/outputs to clear.
 clear_history
 
 #######################################################################
+# methods to print history or any subset thereof
+function print_history(io::IO, indices)
+    for n in indices
```
stevengj: You can just do `for n in intersect(indices, 1:IJulia.n)` and remove most of the special cases below

src/IJulia.jl
```diff
@@ -211,6 +211,50 @@ a subset of cell inputs/outputs to clear.
...
+      if haskey(In, n)
+        print(In[n])
+      end
+    end
+end
+
+# since a range could be huge, intersect it with 1:n first
+print_history{T<:Integer}(io::IO=Base.STDOUT, r::Range{T}=1:n) =
```
stevengj: Just define:
```jl
print_history() = print_history(1:n)
print_history(x) = print_history(STDOUT, x)
```
and then you can remove most of the specialized methods.

dpsanders:
Any particular reason for not just calling the function `history()`?

crstnbr:
I can't find a `history()` function in IJulia. Which function do you mean?

src/IJulia.jl
crstnbr: Yes, will correct it.

dpsanders:
My question is whether history would be a better function name than print_history.

crstnbr:
There is a function `clear_history()` which is why I chose `print_history()`. But `history()` would also be fine with me.

src/IJulia.jl
```diff
@@ -211,6 +211,36 @@ a subset of cell inputs/outputs to clear.
...
+        print(In[n])
+      end
+    end
+end
+history(x) = history(Base.STDOUT, x)
+history() = history(1:n)
+
+# print the last x history entries when argument x is integer
+history(io::IO, x::Integer) =
+    invoke(history, Tuple{IO, Any}, io, n-x:n)
```
stevengj: No need for `invoke` here.

src/magics.jl
```diff
@@ -121,11 +121,11 @@ gui_magic_help(magic::AbstractString, args::AbstractString) = md"""
...
-    access to the input history, is to access the global variable `In`
-    in IJulia.  `In` is a dictionary mapping cell numbers to the inputs.
-    However, IJulia does not currently keep any other history, e.g.
-    it discards input cells that you overwrite.
+    An analogue of the `%history` magic of IPython, which provides
+    access to the input history, is given by IJulia.history(). It is
+    based on the global variable `In` in IJulia. `In` is a dictionary
+    mapping cell numbers to the inputs. However, IJulia does not currently
+    keep any other history, e.g. it discards input cells that you overwrite.
 """
```
stevengj: We should also add this help for `%hist`, since I guess that is a synonym for `%history` in IPython?

src/IJulia.jl
stevengj: Actually, I would define something more general, to match IPython:
```jl
function history(io::IO, indices::AbstractVector{Int})
    for n in sort!(unique(intersect(indices, 1:IJulia.n)))
        ...
    end
end

history(io::IO, x::Union{Integer,AbstractVector{Int}}...) = history(io, vcat(x...))
history(x...) = history(STDOUT, x...)
history(io::IO, x...) = throw(MethodError(history, (x...)))
```
This way you can do e.g. `history(3, 3:7, 9)` 

(I would have used `AbstractVector{T<:Integer}`, but some tricky dispatch issues arose, and it didn't seem worth resolving them.)

stevengj:
Do we want `history(io::IO, r::Regex)`, similar to IPython?

src/magics.jl
crstnbr: Sure, I'll add this.

src/IJulia.jl
crstnbr: I thought it would be nice to have history(x::Integer) to show the latest x elements of IJulia.In. But this is of course different from how IPython's %history behaves. Maybe we should try to stay similar. 

In this spirit, we should, however, stick with `for n in intersect(indices, 1:IJulia.n)` and neither sort nor check for uniqueness, as `%history 4 4 2` in IPython is perfectly valid and gives two times the fourth and the second history entry.

I'll change the rest according to your nice suggestion.

crstnbr:
Doesn't seem to be very useful to me. If we would add it, should it always search the full history?

src/IJulia.jl
stevengj: (We aren't bound to emulate IPython's behavior, of course, but it's a hint about what people have found useful.)

src/IJulia.jl
```diff
@@ -211,6 +211,34 @@ a subset of cell inputs/outputs to clear.
...
+    history([io], [indices])
+
+The `history()` function prints all of the input history stored in
+the running IJulia notebook in a format convenient for copying.
+
+The optional `indices` argument is a collection of indices indicating
+a subset of cell inputs to print.
+
+The optional `io` argument is for specifying an output stream. The default
+is Base.STDOUT.
```
stevengj: Should be
```
`STDOUT`
```

src/IJulia.jl
```diff
@@ -211,6 +211,34 @@ a subset of cell inputs/outputs to clear.
...
+      end
+    end
+end
+
+history(io::IO, x::Union{Integer,AbstractVector{Int}}...) = history(io, vcat(x...))
+history(x...) = history(STDOUT, x...)
+history(io::IO, x...) = throw(MethodError(history, (x...)))
+history() = history(1:n)
+"""
+    history([io], [indices])
```
stevengj: `[indices...]`

src/IJulia.jl
```diff
@@ -211,6 +211,34 @@ a subset of cell inputs/outputs to clear.
...
+
+The `history()` function prints all of the input history stored in
+the running IJulia notebook in a format convenient for copying.
+
+The optional `indices` argument is a collection of indices indicating
```
stevengj: `is one or more indices or collections of indices indicating a subset input cells to print`

stevengj:
Thanks!

crstnbr:
Thank you for helping me. Next time I'll be more efficient :)

PR #508, opened 2017-01-26
Evaluate result in current_module(). Fix #507

### LINKED ISSUE #507: current_module() is now sometimes IJulia
Running this code in a fresh notebook now prints `IJulia` on Julia 0.5. It used to print `Main`:

```julia
immutable Foo end
function Base.show(io::IO, ::MIME"text/html", ::Foo)
    @show current_module()
end
Foo()
```

I've bisected the change to [this commit](https://github.com/JuliaLang/IJulia.jl/commit/3ed9573d517fa62dd3dd56c084d7549ec372c6ec). I have some code that relies on the former behavior. It's admittedly not the cleanest code. Should I fix it not to depend on `current_module()`, or is this a legitimate IJulia bug?

cstjean:
I took @yuyichao's suggestion.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -78,7 +78,7 @@ function error_content(e, bt=catch_backtrace(); backtrace_top::Symbol=:include_s
     evalue = try
         # Peel away one LoadError layer that comes from running include_string on the cell
         isa(e, LoadError) && (e = e.error)
-        sprint((io, e, bt) -> eval(:(showerror($io, $(QuoteNode(e)), $bt, backtrace=false))), e, bt)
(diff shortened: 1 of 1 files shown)


cstjean:
Is there an easy way to test this?

stevengj:
There are maybe other calls to eval that should be fixed, e.g. in `error_content`, for `user_expressions[v] = eval(Main,parse(ex))`.

I guess `include_string` already evaluates in `current_module()`?

cstjean:
> There are maybe other calls to eval that should be fixed, e.g. in error_content, for user_expressions[v] = eval(Main,parse(ex)).

I just fixed those.

Issue #504, opened 2017-01-23
IJulia won't start in Julia 0.6


NickMcNutt:
I upgraded Julia to master a few days ago, and it seems to no longer be compatible with IJulia.  Here is the Jupyter log.

```
[W 12:35:35.833 NotebookApp] All authentication is disabled.  Anyone who can connect to this server will be able to run code.
[I 12:35:35.845 NotebookApp] Serving notebooks from local directory: /Users/Nick/projects/groupsync
[I 12:35:35.845 NotebookApp] 0 active kernels
[I 12:35:35.845 NotebookApp] The Jupyter Notebook is running at: http://localhost:8888/
[I 12:35:35.845 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[I 12:35:40.122 NotebookApp] 302 GET / (::1) 1.05ms
[I 12:35:44.131 NotebookApp] Creating new notebook in
[I 12:35:44.915 NotebookApp] Kernel started: eacff704-2f1d-4893-97bb-e5781d9d6b39
[W 12:35:44.933 NotebookApp] 404 GET /nbextensions/widgets/notebook/js/extension.js?v=20170123123535 (::1) 13.63ms referer=http://localhost:8888/notebooks/Untitled5.ipynb?kernel_name=julia-0.6
Starting kernel event loops.
[E 12:35:50.466 NotebookApp] Bad kernel_info reply
    Traceback (most recent call last):
      File "/usr/local/lib/python2.7/site-packages/notebook/services/kernels/handlers.py", line 164, in _handle_kernel_info_reply
        msg = self.session.deserialize(msg)
      File "/usr/local/lib/python2.7/site-packages/jupyter_client/session.py", line 859, in deserialize
        raise ValueError("Invalid Signature: %r" % signature)
    ValueError: Invalid Signature: '476cdf035239ecaa79ae8fce18ab2a4917d7ca8fcec5b6bddec435102c0f4a9f'
[E 12:35:50.506 NotebookApp] Uncaught exception GET /api/kernels/eacff704-2f1d-4893-97bb-e5781d9d6b39/channels?session_id=1B0A0CDD128C4E58BD36B99AE0E85A60 (::1)
[...]

PR #513, opened 2017-02-12
pre/post execute hooks fail due to world age

adambrewster:
When trying to use PyPlot with IJulia notebook, I get the following error (after fixing https://github.com/JuliaLang/IJulia.jl/issues/504):

```
MethodError: no method matching display_figs()
The applicable method may be too new: running in world age 21524, while current world is 21536.
Closest candidates are:
  display_figs() at /home/adam/.julia/v0.6/PyPlot/src/PyPlot.jl:110 (method too new to be called from this world context.)

Stacktrace:
 [1] execute_request(::ZMQ.Socket, ::IJulia.Msg) at /home/adam/.julia/v0.6/IJulia/src/execute_request.jl:175
 [2] eventloop(::ZMQ.Socket) at /home/adam/.julia/v0.6/IJulia/src/eventloop.jl:8
 [3] (::IJulia.##9#12)() at ./task.jl:335
```

I'm not totally clear how the new world age stuff is supposed to work, but this three line patch gets IJulia + PyPlot back to working on today's master.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -149,7 +149,7 @@ function execute_request(socket, msg)
 
     try
         for hook in preexecute_hooks
-            hook()
+            @eval $hook()
         end
 
         #run the code!
@@ -172,7 +172,7 @@ function execute_request(socket, msg)
         end
 
         for hook in postexecute_hooks
-            hook()
+            @eval $hook()
         end
 
         # flush pending stdio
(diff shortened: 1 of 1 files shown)


darwindarak:
Is it possible to tag a new version that includes this fix?

stevengj:
I just tagged 1.5.0: https://github.com/JuliaLang/METADATA.jl/pull/8937

stevengj:
This should be changed to use `Compat.invokelatest`

PR #521, opened 2017-03-22
Fix a typo in IJulia.jl

achimnol:
* `pretexecute_hooks` must be `preexecute_hooks`.


tkelman:
Was this not tested?

stevengj:
No.  Right now we don't really have a good automated testbed for IJulia.

stevengj:
(It was probably never used because almost always packages add hooks, but never remove them.)

PR #525, opened 2017-04-02
seed explicitly MersenneTwister(0)

rfourquet:
The default constructor `MersenneTwister()` will be deprecated, cf. https://github.com/JuliaLang/julia/pull/16984#issuecomment-290914407.


src/init.jl
```diff
@@ -4,7 +4,7 @@ include(depfile) # generated by Pkg.build("IJulia")
 
 # use our own random seed for msg_id so that we
 # don't alter the user-visible random state (issue #336)
-const IJulia_RNG = MersenneTwister()
+const IJulia_RNG = MersenneTwister(0)
```
stevengj: I think we want `srand(MersenneTwister(0)))`

rfourquet: IIUC, you added this `IJulia_RNG` as a replacement to `GLOBAL_RNG` in `uuid4` to solve an issue, so indeed it seems that it should be randomly seeded as `GLOBAL_RNG` is.

PR #542, opened 2017-05-15
Call message handlers with invokelatest on 0.6-rc1+

JobJob:
This is needed for Interact to work on 0.6

- [x] Fix REQUIRE for Compat after https://github.com/JuliaLang/Compat.jl/pull/359 is merged

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,5 +2,5 @@ julia 0.5
 MbedTLS 0.4.3
 JSON 0.5
 ZMQ 0.3.3
-Compat 0.9.5
+Compat 0.25.1
 Conda 0.1.5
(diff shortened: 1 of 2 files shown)


stevengj:
Compat v0.25.0 adds `invokelatest` (https://github.com/JuliaLang/Compat.jl/pull/352), so we can just use `invokelatest` unconditionally if you bump the Compat requirement.  (Similarly for other places where we currently call `eval`.)

JobJob:
Ok, do you mind if I close this and leave that to you? Spent a long time down the rabbit hole fixing Reactive issues.

JobJob:
Ok, updated this. Waiting on https://github.com/JuliaLang/Compat.jl/pull/359 (and a tag) so we can update the REQUIRE, otherwise this won't work on julia 0.5

stevengj:
Thanks!

JobJob:
Updated. Oh https://github.com/JuliaLang/METADATA.jl/pull/9334 wasn't merged yet in Metadata, causing test failures.

PR #560, opened 2017-06-25
	modified:   deps/build.jl

ajalexei:
On linux systems, try "jupyter-notebook" to get the
Jupyter version in case "jupyter" command fails. This is the
case for the Debian distribution.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -23,6 +23,9 @@ end
 
 jupyter = get(ENV, "JUPYTER", isfile("JUPYTER") ? readchomp("JUPYTER") : is_linux() ? "jupyter" : "")
 jupyter_vers = isempty(jupyter) ? v"0.0" : prog_version(jupyter)
+if (jupyter_vers == v"0.0") && is_linux()                                       # some Linux distributions (Debian) use jupyter-notebook to launch Jupyter
+    jupyter_vers = prog_version(jupyter * "-notebook")
+end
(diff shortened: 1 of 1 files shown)


stevengj:
Presumably, this isn't sufficient to get it to work, since you also have to fix the kernelspec command and the notebook command stored in deps.jl.

ajalexei:
It does work on my Debian box. The trick seems to be that the notebook and kernelspec commands are first set to "jupyter notebook|kernelspec". if that fails, then the next candidates on unices are "jupyter-notebook|kernelspec" - precisely the ones that Debian uses. So the only problem was to set the "jupyter_vers" correctly for that case.

deps/build.jl
```diff
@@ -23,6 +23,9 @@ end
 
 jupyter = get(ENV, "JUPYTER", isfile("JUPYTER") ? readchomp("JUPYTER") : is_linux() ? "jupyter" : "")
 jupyter_vers = isempty(jupyter) ? v"0.0" : prog_version(jupyter)
+if (jupyter_vers == v"0.0") && is_linux()                                       # some Linux distributions (Debian) use jupyter-notebook to launch Jupyter
```
stevengj: I would get rid of the `is_linux()` test, since presumably jupyter may be installed as `jupyter-notebook` on other systems too.

ajalexei:
Indeed, that seems to be ok, so I updated the code. I cannot test it though, as I have no access to a Windows machine with Julia installed.

Issue #562, opened 2017-06-27
IJulia status replies come in really slowly


JMurph2015:
I was benchmarking the performance of a jupyter_client backend, and I hypothesized that the fastest kernel for execution overhead times would be IJulia given that it's native Julia and not a IPython wrapper.  IJulia is slightly faster when it comes to sending 'execute_reply's, but it is an order of magnitude (~70ms or much more sometimes, >400ms) slower than IPython to publish the 'idle' status on IOPub.  

This is important because my client needs to wait for the 'idle' message so that it knows when all stdout and stderr streams have been pushed.  Thus, IJulia is roughly an order of magnitude slower or more when it comes to real world performance for me.

I don't really know how to go about fixing this, but I thought it was worth posting an issue to raise awareness.

yuyichao:
https://github.com/JuliaLang/IJulia.jl/pull/491

JMurph2015:
This is pretty evil behavior if you ask me.  

Such arbitrary delays are not a part of the kernel protocol and degrade the interoperability of the kernels (due to my application now having to specially initialize IJulia kernels differently than others so that things don't slow to an absolute crawl).  

A Signal interface on which processes can require Julia to hold is one thing, but the stopgap of just setting an arbitrary (and relatively high) delay on all 'idle' messages is pretty unsightly (even if it is resettable by user processes).  It should always be the bare minimum the currently running code needs (aka an async hold condition).

JMurph2015:
For context, your default delay slows IJulia from 1.5x faster than IPython to 42.3x slower than IPython, aka slows Julia down by a factor of 63.5.  That's a _nearly two orders of magnitude_ slowdown.

JMurph2015:
Just an update, I created a fork and made a channel interface (using channels with size 1 as semaphores).  If you want to hold the 'idle' reply, you just register a semaphore and fulfill it sometime between t=0 and t=timeout.  When you are done you just call IJulia.remove_hold_channel with the instance, or you close the channel and it will automatically be culled.

koffie:
I think this is indeed an important thing to fix. Nice to see that you made a pull request for this. This timeout of 0.5 seconds is really ridiculous, it is so long that it is even noticeable by the enduser. Just in case someone comes here before the pull request is merged: just do the following in your notebook to make it feel snappy again.
```
import IJulia
IJulia.set_idle_delay(0.01)
```

stevengj:
[...]

PR #563, opened 2017-06-28
Improve idle holding behavior through the use of a channel-based interface

JMurph2015:
This addresses #562 by creating a channel interface where user code can register channels as semaphores and hold IJulia from firing the 'idle' message.  Also sets the timeout to 0 by default.  Interact.jl compatibility as discussed in #491 is also checked for and naively waited on so that it works until Interact.jl updates to use the channel interface to hold until necessary.  Hope this helps.  

If a new type or something is needed, that works for me, I just used the most basic level-trigger that Julia has in the base docs.  

Also could use assistance with an error path in case one of the semaphores times out (aka some way to indicate to the user which semaphore was causing issues and where that semaphore came from)

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 MbedTLS 0.4.3
 JSON 0.5
-ZMQ 0.3.3
+ZMQ 0.6.0
 Compat 0.41.0
 Conda 0.1.5
--- a/src/heartbeat.jl
+++ b/src/heartbeat.jl
@@ -1,24 +1,14 @@
 # Spawn a thread (using pthreads on Unix/OSX and Windows threads on Windows)
 # to implement the "heartbeat" message channel, which is just a ZMQ
-# socket that echoes all messages.  This is implemented with the zmq_device
+# socket that echoes all messages.  This is implemented with the zmq_proxy
 # call in libzmq, which simply blocks forever, so the usual lack of
 # thread safety in Julia should not be an issue here.
 
-const libzmq = isdefined(ZMQ, :libzmq) ? ZMQ.libzmq : ZMQ.zmq
-
 # entry point for new thread
-if ZMQ.version ≥ v"3.2.1" # always true once we require a newer ZMQ.jl
-    function heartbeat_thread(sock::Ptr{Void})
-        ccall((:zmq_proxy,libzmq), Cint, (Ptr{Void}, Ptr{Void}, Ptr{Void}),
-              sock, sock, C_NULL)
-        nothing
-    end
-else
-    function heartbeat_thread(sock::Ptr{Void})
-        ccall((:zmq_device,libzmq), Cint, (Cint, Ptr{Void}, Ptr{Void}),
-              ZMQ.QUEUE, sock, sock)
-        nothing
-    end
+function heartbeat_thread(sock::Ptr{Void})
+    ccall((:zmq_proxy,ZMQ.libzmq), Cint, (Ptr{Void}, Ptr{Void}, Ptr{Void}),
+          sock, sock, C_NULL)
+    nothing
 end
 
 function start_heartbeat(sock)


JMurph2015:
@yuyichao @JobJob @stevengj

JMurph2015:
I looked at the errors on the appveyor tests, they don't appear to be related to my code, but rather something about ZMQ not being installed correctly.  Take a look and let me know if that's indeed the case or if I need to change something to play more nicely.

stevengj:
Yeah, the appveyor tests have been broken for a while and I haven't gotten around to fixing them.

src/eventloop.jl
```diff
@@ -18,7 +18,15 @@ function eventloop(socket)
                 end
             finally
                 @async begin
-                    sleep(idle_delay[])
+                    if check_interact_loaded() && idle_delay[] < interact_min_delay
```
stevengj: I don't like specifically checking for Interact.   Would be better to patch Interact to use the same mechanism for controlling the idle delay, or simply to get rid of the idle delay entirely in favor of channels or whatever.

JMurph2015: So this was intended to be a deprecation sort of warning, where for the time being, I won't break Interact.jl, and this PR can get processed (as I haven't even looked at Interact's source to patch it.)

I promise someone can remove it a version or two down the line.

src/eventloop.jl
```diff
@@ -52,7 +60,72 @@ function waitloop()
...
+    put!(notification, true)
+end
+
+#= Interact.jl compatibility code
+    This ensures the existing fix continues to work with Interact.jl until
+    they can update to using the new semaphore or channel interface.
+=#
+const interact_min_delay = 0.25
+function check_interact_loaded()
+    return isdefined(:Interact)
```
stevengj: `isdefined(Main, :Interact)`, probably.

src/eventloop.jl
stevengj: The probably just
```jl
if isdefined(Main, :Interact) # eliminate this once Interact is updated
    sleep(0.25)
end
```
and get rid of `idle_delay[]`?  As far as I can tell, no packages are actually using `set_idle_delay`, so we can eliminate it.

src/eventloop.jl
JMurph2015: I'm okay with that.  It is a little more general to say isdefined(:Interact), but if you want to restrict it to Main that's okay too.  The penalty for a false positive here isn't huge, at least in the sense that it's unlikely someone would define Interact and care about response performance.

JMurph2015:
Apart from the error paths, a future issue that I would be willing to work on (but also could use assistance with) is that I don't have a solid testing scheme for this yet, so I would like to add some test cases to /tests to ensure continuing functionality.

src/eventloop.jl
JMurph2015: I'm alright with that, I can update my branch to eliminate the idle delay functionality.

JobJob:
Notwithstanding the hyperbole in the original issue, which I could only laugh at, this seems good!

Comments:
1. IJulia will process more `execute_request` messages (and the frontend will send them, e.g. if you run multiple cells) even if the status:idle hasn't been sent for a previous cell, so actual performance in the notebook, i.e. the vast majority of usage (~= all use cases besides yours 😄 ), isn't degraded by the existing behaviour. You can test this out by setting the idle delay really high, to say 10 seconds, and running multiple cells in a notebook.
1. #491 is a workaround for what I think can be fairly regarded as a bug in the frontend. We would happily send the idle asap, but if we do, output can get lost if there's an intervening comm_msg. Having the delay is a super simple solution that may help other packages that also use custom `comm_msg`s without them having to make any changes, with no real downside, or so it seemed. Nonetheless, this PR, whilst more complex, seems to be a neater and more powerful solution, so thanks!

I'll hopefully get a chance to look at it in more detail (and try it out) over the weekend. I'll probably make the necessary changes to Interact myself.

JMurph2015:
I will admit I was probably a little more annoyed than was really warranted in the original issue, but to me it was just that a large portion of the reason Julia exists is for performance concerns, so to me when I asked myself "which should be the fastest Jupyter kernel?" I immediately thought of IJulia.  So then to learn it was two orders of magnitude slower than it really had to be, it was quite a letdown considering pretty much everything in Julia is by default orders of magnitude faster than a language like Python.

JMurph2015:
@stevengj How do you feel about the modifications thus far?  Are any other revisions in order?

JobJob:
These are my thoughts after taking a closer look at this.

I see the cases this PR might touch are:

1) Regular case: normal synchronous code execution and output
2) Long running asynchronous tasks where comm/other messages are sent as well as output
3) @JMurph2015's case: wants to receive the idle as soon as possible
4) Interact case: output truncation in "slightly asynchronous" cases (output happens from an asynchronous task which runs very soon after the cell code is run), where other messages are also sent, such as the comm messages of Interact*

`*` when a new, e.g. comm, message is sent after the idle is received, Jupyter's javascript in the browser removes the output display handlers, so stdout/stderr stream messages aren't displayed.

How this PR affects these cases:
1. No effect AFAICT
2. I was hoping it would help here, and that this would be the payoff, but it doesn't at present, because of the `hold_channels_timeout_default`
3. Helps, although currently case 3 is fine as long as you call `IJulia.set_idle_delay(0.0)`

Regarding case 4, this is currently handled with the default 0.5s delay. Handling all the code paths required to correctly use the channels in this PR seems like it might get complicated.

To me it seems like just calling `IJulia.set_idle_delay(0.0)` to handle case 3, is going to be simpler than manually working out where Interact/Reactive (and possibly other packages) need to hold and release the semaphores, to handle case 4 (e.g. PlotlyJS uses it's own custom comm channel and may possibly run into it). To me it seems case 4 affects more people, and might be difficult to get right using the mechanism in this PR, and case 3 has a fairly simple workaround.

@JMurph2015 - is there a reason why you're so averse to using `IJulia.set_idle_delay(0.0)`, what if we set it up to use an environment variable - would that make your life easier? e.g. put `"IJULIA_IDLE_DELAY" in keys(ENV) && IJulia.set_idle_delay(ENV["IJULIA_IDLE_DELAY"])` in kernel.jl ?

However, if this PR could be be adapted so that the default delay was left at 0.5s, but provided a reliable mechanism in case 2, for indefinitely delaying the status:idle from being sent, then I'd be on board.

Overall, even though this PR is neat work, it may put a burden of effort on other packages, and I guess I still don't really see a major downside to delaying the idle by 0.5s by default. Nonetheless, a mechanism to indefinitely delay the idle would probably be useful for some cases. But I completely understand if you don't feel like continuing down that path if it doesn't serve your aims.

Having said all that, it is just, like, my opinion 😄 and I'm open to counter points.

JMurph2015:
@JobJob I'm mostly opposed to the `IJulia.set_idle_delay()` on principle.  Sure one could work around it without a ridiculous burden, but every other Jupyter kernel behaves as if `IJulia.set_idle_delay(0.0)` was the default (or if this PR was accepted).  Furthermore, Julia being a language that functionality just doesn't get implemented if it compromises speed, it would make sense that as much as possible we would push possible slowdowns as far down the stack as possible ( remember IJulia is installed on pretty every Julia install ).

Therefore, if Interact or Reactive need to have some funky delay mechanism, I have no problems making a hook for them to do that, but I do have a problem with IJulia doing it for them.  In fact this new interface could use a hook for IJulia to signal a 'idle' message sent event, so that Interact could just set up an async routine that waited for that signal, then waited 0.5s, then released IJulia, wash-rinse-repeat.

Secondarily, I'm opposed on principle to delay mechanisms anywhere in code where it is just trying to approximate the runtime of an async routine.  There is definitely _something_ that you are waiting on, so the rub is just finding it and being able to put essentially a callback on it.  Interact should have some "comm-signal" pipeline, and the trick is to just register the semaphore somewhere in the setup process, clear it at the beginning of the pipeline, then trigger it at the end of that "comm-signal" pipeline.  Now, I did take a look at the Interact code, and I will say this looks tricky without being able to hook into some of the code outside of the ijulia.jl file there.

MIT has a hacking rule "Brute force is the last resort of the incompetent", and I generally apply that to code as well. [http://mitadmissions.org/blogs/entry/practice_safe_hacking](http://mitadmissions.org/blogs/entry/practice_safe_hacking), which leads us to...
Three takeaways from my side here:
1.  It may be somewhat harder (and in some cases legitimately difficult) to implement semaphores, and it should be made serviceable to just fire off a delay async routine. These async routines would live in other packages' code, and Interact could use something like this until we find the exact points where the semaphore can be effectively used.
2. The semaphore system, while more complex, is the best at keeping IJulia fast and elegant, which is somewhat more important than keeping Interact's and React's hands clean.
3. Considering the pains taken to make other parts of the Julia ecosystem as fast as possible, the speedup and elegance of this is well worth our effort even if we have to rewrite a couple files or add some callbacks to some functions.

PS Interact and React could've actually implemented stuff on their end in the first place that would make sure that their main threads waited for side-jobs to finish in IJulia contexts which would have eliminated this issue before the days of the `IJulia.set_idle_delay()`

JMurph2015:
Also I just processed what you meant about use case (2).  Yeah, so I wouldn't mind some way to indefinitely hold IJulia, but it's just that it would really mess things up if someone abused it, because then things go very badly.  I'm open to suggestions on how long that hold_channels_timeout_default should be, and also could be convinced to open up an interface to changing that value.

JobJob:
I appreciate, and share, your aim to improve code quality across the Julia ecosystem. I guess this issue comes down to the trade-offs that are acceptable to people doing the work. My view is IJulia is fast - it just doesn't send the idle fast and I don't see that as a major problem (c.f. https://github.com/jupyter/notebook/issues/518).

> "Brute force is the last resort of the incompetent"

I also think there are other maxims of programming that apply:
1) premature optimisation is the root of all evil (in this case I see it as more unnecessary than premature)
2) done is better than perfect
3) real artists ship

😄 

Anyway, I personally am ok with the status quo, but if you want to finish this up and make a PR to Interact - I think probably just adding a hold, then async release of the semaphore after, e.g., a 0.5s delay, in all the `handle_msg` methods in https://github.com/JuliaGizmos/Interact.jl/blob/master/src/IJulia/handle_msg.jl should (inelegantly but simply) work - then great!

> I wouldn't mind some way to indefinitely hold IJulia, but it's just that it would really mess things up if someone abused it, because then things go very badly

I actually don't think it would be that bad - not sending idles is not a big deal is it? If someone is using this system then it's up to them to get it right IMO. I.e. I don't think there should be a timeout at all, which would enable people with arbitrarily long running processes (and other intervening messages) to avoid losing their output.

JMurph2015:
I'll add a hook so that packages can listen in on when IJulia is sending 'idle' signals.  Then I'll just add an `@async` task to Interact.  I think that is the simplest right now, and makes it possible for other packages architect around channels in the future (it's simpler to include this sort of thing on the ground floor of a package.)

Secondarily, has anyone seen the Summer of Code project on Interact [https://summerofcode.withgoogle.com/projects/#5415076583440384](https://summerofcode.withgoogle.com/projects/#5415076583440384)?  Could we have them consider this from the get-go on that project?

JobJob:
> I'll add a hook so that packages can listen in on when IJulia is sending 'idle' signals. Then I'll just add an `@async` task to Interact.

Sounds good!

> Secondarily, has anyone seen the Summer of Code project on Interact https://summerofcode.withgoogle.com/projects/#5415076583440384? Could we have them consider this from the get-go on that project?

Yep, I'm intimately familiar that work 😉 All the IJulia related stuff is happening in [WebIO](https://github.com/JuliaGizmos/WebIO.jl) now.

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+# Interact.jl band-aid compatibility code.
+const interact_min_delay = 0.25
+
+# Channel constants
+const hold_channels = Array{Channel{Bool},1}()
+const hold_channels_timeout_default = 3
+
+function add_hold_channel(channel::Channel{Bool})
+    if !(channel in hold_channels)
+        hold_channels = [hold_channels..., channel]
```
stevengj: use `push!`

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+    end
+end
+
+function remove_hold_channel(channel::Channel{Bool})
+    if channel in hold_channels
+        filter!(e->e∉[channel],hold_channels)
```
stevengj: Why not `filter!(e->e != channel ,hold_channels)`?  And the `channel in hold_channels` check seems redundant.

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+    See wait_multi_with_timeout()
+=#
+# Interact.jl band-aid compatibility code.
+const interact_min_delay = 0.25
+
+# Channel constants
+const hold_channels = Array{Channel{Bool},1}()
+const hold_channels_timeout_default = 3
+
+function add_hold_channel(channel::Channel{Bool})
```
stevengj: Needs a docstring, similarly below, since non-IJulia modules are expected to use this.

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+    reach into their channel state manually unless there is a
+    timeout, in which case a temporary variable is put onto the channel
+    buffer and removed after the wait process.
+    See wait_multi_with_timeout()
+=#
+# Interact.jl band-aid compatibility code.
+const interact_min_delay = 0.25
+
+# Channel constants
+const hold_channels = Array{Channel{Bool},1}()
```
stevengj: `hold_channels` is not very descriptive.  Maybe `idlechecks`, `add_idlecheck`, etc?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+        that don't come through before the timeout.  Doing this because someone
+        will inevitably not fulfill the semaphores that they register, and
+        thus this will take care of us, at least for the short term.
+    =#
+    @async begin
+        sleep(timeout)
+        idxs_to_reset = []
+        for i in eachindex(channels)
+            if length(channels[i].data) == 0
+                put!(false)
```
stevengj: missing a channel argument?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+                idxs_to_reset = [idxs_to_reset..., i]
+            end
+        end
+        wait(notification)
```
stevengj: This seems like a potential race condition.   What if `wait_multi_with_timeout` completes (1) and calls `put!(notification, true)`, then `wait_multi_with_timeout` is called again quickly (1), and *then* the `wait(notification)` on the *original* (1) call wakes up and resets the channels?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
-const idle_delay = Ref(0.5)
-function set_idle_delay(delay_secs::Float64=0.5)
-    idle_delay[] = delay_secs
+#=
+    Hold Channel Interface
+    This allows user processes to register and deregister
+    channels that allow user code to hold IJulia from sending
+    the idle status message until signaled.  This would have
+    been prettier if there were level-triggered events in
+    Julia base, but this will do okay for now.
```
stevengj: From the `Condition` documentation:
```
  Create an edge-triggered event source that tasks can wait for. Tasks that
  call wait on a Condition are suspended and queued. Tasks are woken up when
  notify is later called on the Condition. Edge triggering means that only
  tasks waiting at the time notify is called can be woken up. For
  level-triggered notifications, you must keep extra state to keep track of
  whether a notification has happened. The Channel type does this, and so can
  be used for level-triggered events.
```

stevengj: It's not clear to me why we're using `Channel` and not `Condition` here.   If we used `Condition`, then tasks could just call `notify()` when they are idle, no?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+        idxs_to_reset = []
+        for i in eachindex(channels)
+            if length(channels[i].data) == 0
+                put!(false)
+                idxs_to_reset = [idxs_to_reset..., i]
+            end
+        end
+        wait(notification)
+        for i in idxs_to_reset
+            take!(channels[i])
```
stevengj: we don't look at the value of the `channel`?

src/eventloop.jl
stevengj: Do we require that the channel is unbuffered?  We can check for that with `Base.isbuffered`

src/eventloop.jl
stevengj: Oh, I see, it is for level triggering that we use unbuffered Channels.  But then what do you mean about Julia not having level triggering in base?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+    end
+end
+
+function remove_hold_channel(channel::Channel{Bool})
+    if channel in hold_channels
+        filter!(e->e∉[channel],hold_channels)
+    end
 end
+
+function wait_multi_with_timeout(channels::Array{Channel{Bool},1}, timeout::Number)
```
stevengj: @yuyichao, is there a better way to do this in Base?  I feel like there must be a cleaner way to do this sort of thing.

stevengj:
I feel like there should be a single semaphore that user code can increment when it wants Julia to wait before idling, and decrement when they are ready to idle.  Why do we need multiple channels?

src/eventloop.jl
```diff
@@ -52,7 +57,63 @@ function waitloop()
...
+    notification = Channel{Bool}(1)
+    #=
+        This async routine will manually fulfill any channels (aka semaphores)
+        that don't come through before the timeout.  Doing this because someone
+        will inevitably not fulfill the semaphores that they register, and
+        thus this will take care of us, at least for the short term.
```
stevengj: I'm not clear on what the timeout is for.   User code is free to do infinite loops, or sleep forever, so why would we try to catch bugs where they failr to release the semaphore?

JMurph2015:
Uh, so oops on my git skills, I had some stuff I was testing, forgot it would push to this PR when I pushed to my fork, and here we are (after some squashing that didn't really even help).

I can also look into the Reentrant lock business.  I'm kinda partial to my Event type, for obvious reasons, but my favorite reason is because I solved a race condition of code listening for an IJulia "ready to send" signal and IJulia starting the waiting process on their hold (possibly before it was cleared), so I tossed a "persistent" field on the type to designate which should be cleared on every loop (by IJulia) and which should be left alone.  That way people can either hold in reaction to IJulia being ready or just hold whenever they want, like at the start and end of an async routine.

src/eventloop.jl
JMurph2015: A lot of the time when a language/library supports a Condition type, right next to it in the docs will be an Event type that is just for level triggering.  Unbuffered channels work for it, but they aren't really made for that purpose.

src/eventloop.jl
JMurph2015: I think here I was thinking that if one wanted to extend this method to a batch remove, then this was more prepared for that, but I guess that's somewhat premature optimization.

JMurph2015:
Also, I'm not too averse to dropping the timeout, that was mainly in the principle of isolating potential bugs from say an imported package from holding up functionality of the rest of the running code (via this feature at least).  Although overall, the odds of that seem low, and it's not the end of the world.

stevengj:
Putting aside the timeout, can't we use a simpler semaphore-like implementation?  A task increments an idle counter when it is not idle, and decrements it when it is, and the main task waits for the idle counter to hit zero before sending the idle message.  i.e. something like:
```jl
const busycount = Ref(0)
const idlecond = Condition()
function busy()
    busycount[] += 1
end
function idle()
    busycount[] -= 1
    notify(idlecond)
end
function waitidle()
    while busycount[] > 0
        wait(idlecond)
    end
end
```
Then tasks that want IJulia to wait before idling would call `IJulia.busy()` and then `IJulia.idle()` when they are done.

JMurph2015:
The only problem I have with a straight counter is that it doesn't distinguish by "user".  For example
```
Task 1:
IJulia.busy()
...
IJulia.idle()
Task 2:
IJulia.busy()
...
IJulia.idle()
```
is equivalent to
```
Task 1:
IJulia.busy()
IJulia.idle()
IJulia.idle()
Task 2:
IJulia.busy()
```
So maybe some registered increment/decrement would be good?

stevengj:
If task 1 calls `busy` once and `idle` twice, that is a bug; I'm not sure we need to protect against this.   As long as tasks correctly match the count of busy and idle calls, we don't need to distinguish the count by user.

stevengj:
(This is a low level interface that relatively few packages, possibly only Interact, will be using; it's not like this is an API for end users.   And if Interact is buggy, it already has lots of ways to screw up IJulia.)

JMurph2015:
Okay, that's fair, I'll do some refactoring tonight.

JMurph2015:
So I took it a touch more complicated than your example, just so that I could overload the Base.wait function and use that.  It's just a type with some barebones functions defined for it, namely Base.+, Base.-, wait, and notify.  I think it adds a nice layer of abstraction where, in the loop, in busy(), and in idle(), you use it just like a counter, but in the loop's wait() call and in the wait_for_ready() it behaves like a Condition.

stevengj:
Defining a type and overloading `+` etcetera seems like overkill to me for a type that is used in literally one place.

My suggestion requires exactly 15 lines of code (14 lines of definitions and one call to `waitidle()`), whereas you are adding more than double that.

JMurph2015:
I just liked using the Base.wait function. :/
I can either pull the overrides on + and -, and make them custom methods, or pull out that type entirely...
Also I did miss that your wait_idle probably doesn't actually block anything except the current green thread.  (Sorry, I've been using Python 3.6 too much recently, and its native async is substantially more verbose)
We still have to accomodate a hook for listening in on the loop so that this sort of workaround is tractable
``` julia
@async begin
    while true
        IJulia.wait_for_idle_ready()
        IJulia.busy()
        sleep(0.5)
        IJulia.idle()
    end
end
```
This was the proposed short term solution for Interact.jl-side compatibility, and having two semaphores kinda justified the type in my head, even if it wasn't going to save many lines.

stevengj:
I don't see that it's worth implementing a timeout.  Just accept that if a package is buggy and calls `busy` but never calls `idle` that it will hang IJulia.   There's no point in protecting against this, since if a package just calls `sleep(1e10)` it will also hang IJulia.

I think IJulia's execute function should just replace `sleep(idle_delay[])` with
```jl
sleep(idle_delay[])
wait_idle()
```
and then at some point after Interact is updated we remove the `sleep` line

JMurph2015:
So the "quick fix" @JobJob and I were talking about is that Interact should just have an async process that would hold IJulia for some arbitrary amount of time, every time IJulia is otherwise ready to send an "idle" message.  Therefore, that means one needs some way of knowing that IJulia is ready for this rigamarole so that they can start something like a half second timer, after which they release IJulia.  

This exactly mimics current behavior (at least from Interact's point of view), but moves the burden of deciding how long to wait to Interact.  That is why there needs to be some facility for processes to "listen in on the loop" (IJulia's eventloop) so that they know when to start timers (or other things potentially).  

_The timeout functionality has been entirely removed at this point._

JobJob:
Thanks @JMurph2015 for your good work on this and keeping Interact's workload in mind. I actually had a chance to look into where Interact could use the semaphore today.

It would need to call `IJulia.busy()` the line above here: https://github.com/JuliaGizmos/Interact.jl/blob/master/src/IJulia/setup.jl#L54

And then run this:
```
@async begin
    sleep(1.0) # 1.0 to be safe
    IJulia.idle()
end
```
the line below here: https://github.com/JuliaGizmos/Interact.jl/blob/master/src/IJulia/setup.jl#L61
The cases for Interact are surprisingly hairy, without the above (or the current delay) the following code sometimes truncates the error message:
```
p = Signal(0.0)
push!(p, "1") #error because "1" is a String and p is a Float64 Signal
p
```
If we don't use the async sleep, we could handle this case by putting the `IJulia.idle()` call in the Reactive event loop along with a test to see if there are more Reactive events to process, but that might not handle some other cases too well...

Anyway, @JMurph2015, your solution addressed our previous concerns more thoroughly, and if I hadn't spent a while looking into it today (which I probably shouldn't have, got a lot on my plate), the listening in on the idle workaround would have been very helpful as a stop-gap for Interact, so thank you, and sorry I have been a little too busy to look into this more.

I guess the above changes to Interact will work with both of the above solutions (I presume - just quickly tested with @stevengj's and it seemed to work). So regardless of which is decided, I'm happy if any Interact specific code is removed. I'll be able to get the required changes into Interact by mid next week.

I do think the simple option for other packages offered by the code in [this comment](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-314825206) is a nice pro. Whether it justifies the extra complexity (given that maybe no-one will ever use it), I'm not sure. Happy to defer to the acuity of @stevengj's judgement.

Just FYI to you both, probably no need to address it in this PR, but as it stands the code will send multiple idle messages when the triggering `IJulia.idle` is called, which isn't a problem afaict, just thought to let you both know.

Ok, thanks again for your efforts to improve things and reduce the burden on Interact to conform @JMurph2015. Could be nice to put your LevelTrigger stuff into a package.

stevengj:
@JobJob, if Interact does `sleep(1); IJulia.idle()` then doesn't it defeat the whole purpose of this PR by imposing a fixed delay on every idle status when Interact is loaded?

I thought that the point of this PR is that Interact knows when it is safe to idle.

JMurph2015:
So the idea, was that Interact _should_ know when it is safe, but it was going to take some work to make sure.  

There were two short term fixes that would move IJulia from just arbitrary delays that we have come up with.  The first was to just check if Interact was loaded and wait, and the second was to change Interact to use a delay itself (via the new interface).  However Interact can't do anything until IJulia changes its interface, and even then it will take a similar process of discussing and such to get a final solution done.  Not to mention it is getting completely restructured this summer.

Thus, both the short-short term*, the short term, and the long term solutions are all in there.  It's not my favorite thing either, but generally breaking things isn't good for a PR's chances to get approved.

short-short term:  Before Interact has an PR and update to support using semaphores at all
short term: Before we find the best way to make sure Interact holds IJulia properly
long term: IJulia doesn't have arbitrary delays (at least in our code) and Interact knows how to properly trigger the semaphores.

JMurph2015:
Oops, I didn't read @JobJob 's last comment, so I deleted my last comment saying pretty much what he said.

JMurph2015:
Briefly reading Interact's code, I thought it would be safest to overload a method or two on their base types (like Comm?) such that pushing to them etc. was automatically padded with IJulia hold and release code.  But I could be totally wrong there.  But, I didn't see any automatic showstoppers that would lead to Interact not knowing where to release.

stevengj:
@shashi, can you comment on this?

JobJob:
> @JobJob, if Interact does sleep(1); IJulia.idle() then doesn't it defeat the whole purpose of this PR by imposing a fixed delay on every idle status when Interact is loaded?

No, the `init_comm` method where we'd put the idle delay (links to code in previous comment) only gets called when a Signal is displayed, either through a call to `display` or as a result of a cell. So the sleep before idle will only happen in those cases, certainly not on every cell execution once Interact is loaded.
---
tl;dr of the below: very boring information about the intricacies of Interact and Jupyter stream messages. I guess we should merge a simple version of this PR, since it might help, and shouldn't harm. 

Some more info about the scenario: when this code runs
```
p = Signal(0.0)
push!(p, "1") #error because "1" is a String and p is a Float64 Signal
p
```
Here's some notes on how it plays out:
1. javascript(js): send execute_msg
2. Julia(JL): IJulia: run the code, note the `Reactive.push!(p,"1")` is async, and that error message originates from Reactive.
3. JL: `metadata(x::Signal)` ([defined here](https://github.com/JuliaGizmos/Interact.jl/blob/master/src/IJulia/setup.jl#L89)) called as part of displaying the Signal as result of cell
    1. calls JL:`init_comm(x::Signal)`
4. JL: IJulia sends result of cell
5. JL: send idle (exec req)
    1. js: receipt of idle after an exec request means Jupyter will no longer append stream (STDOUT/ERR) messages to the cell when the stream message explicitly specifies the parent_header associated with that cell's last execute request. 
        1. Subsequent stream messages would normally still be displayed under the last cell executed, but this mechanism is interrupted by the interceding comm message - details below.
6. js: Signal value being displayed causes Interact/setup/ijulia.js to send a "subscribe_mime" custom comm message: tells the Julia side of Interact which mimetype Jupyter chose to display the value with, so it knows which mimetype to send to update the value with when it changes.
    1. js: the sending of comm message causes Jupyter to overwrite its last sent message (the execute_request), and it stores a mapping between the last sent message and the display area to append output to, essentially stopping any future comm messages from being displayed. Annoying, but true.
7. JL: receive the "subscribe_mime" msg, processed by handle_subscriptions (sub)
8. JL: send idle (subscribe_mime req)
9. JL: possibly still sending the error message as a stderr stream message

After this PR Julia can delay the idle by calling `IJulia.busy()` at step 3.i, delaying the idle of step 5, and call `IJulia.idle()` with a delay after step 7, so that 8 happens after  9.

I personally don't see where Interact can decide it's ok to (immediately) send the idle. You can look at it as Interact (or any package that initiates comm messages in the browser after the execute message) "breaking" IO in Jupyter (thanks to Jupyter's annoying behaviour) in step 6, so we need to find a workaround. 

The global delay was a perfectly fine solution, and part of me feels this whole PR was an over-engineered solution to a problem that barely existed that has taken up a lot of time for little immediate gain. But we're here now and I think this PR can actually help Interact under certain timings of the above (certainly increasing the delay in that particular case to 1.0s may help - though I suppose we could've done that with `set_idle_delay`). It may also prove useful for other packages in future that wish to get their stream messages under the correct cell by explicitly setting the parent_header and delaying the idle.

stevengj:
> note the `Reactive.push!(p,"1")` is async, and that error message originates from Reactive.

I don't see why the error message would have to be async.  Why doesn't `push!` throw an error message?  i.e. why isn't `push!` defined something like:
```jl
push!{T}(s::Signal{T}, x::T) = .... whatever ...
push!{T}(s::Signal{T}, x) = push!(s, convert(T, x)::T)
```
in which case `convert` will throw an exception immediately?

JobJob:
In that case, yes the error could be sync, but it's just a toy example. There can be many errors that result from the value propagating through the signal graph. e.g.
```
using Interact
s = Signal(1)
m = map(sqrt, s)
push!(s, -1)
s
```

stevengj:
Maybe Interact should catch the exception and send a Jupyter [`error` message](http://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-errors) instead of writing to `STDERR`?

JobJob:
The exception is caught in Reactive [here](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L344), asynchronously, and the error message is written to STDERR by the default error handling function [here](https://github.com/JuliaGizmos/Reactive.jl/blob/master/src/core.jl#L356). What happens in the front end when you send this message?

If you define a custom error handling function to `push!(s, -1; onerror=custom_err_fn)` - you can see the details with `?push!` so you could easily try sending the error.

I'm now thinking possibly a better solution is for Interact to assume Jupyter is going to use the richest mimetype, i.e. guess the mimetype, and have Interact's ijulia.js only send that "subscribe_mime" comm if it got its guess wrong, I bet that would be rare.

stevengj:
> What happens in the front end when you send the error message?

It does exactly what you want.  Because the error message includes the triggering execute_msg in its header, then the error gets printed in whatever cell that execute_msg came from.

For example, I tried running the following code in IJulia, which waits for 5 seconds and then sends an error message.
```jl
cur_execute_msg = IJulia.execute_msg
@async begin
    sleep(5)
    try
        error("foo")
    catch e
        IJulia.send_ipython(IJulia.publish[], IJulia.msg_pub(cur_execute_msg, "error", IJulia.error_content(e)))
    end
end
```
Even if I execute another cell before the error message is sent, the error message is still attached to the original cell:
image

JobJob:
I found if I add an `IJulia.set_idle_delay(0.0)` at the top, it doesn't work on second and subsequent runs. It does work though if you leave the idle at the default 0.5 - probably because by the time Jupyter receives the idle (after the 0.5 sec delay), the next cell has executed so it associated that idle with the second cell, and cleans up its output area handling rather than the first cell's. 

I also found the behaviour to be the same replacing that error sending with:
```
IJulia.set_cur_msg(cur_execute_msg)
println(STDERR, "he's not the messiah, he's a very naughty boy...")
```
Strange that it works on the first run, but I didn't expect any output to show after the idle is sent, maybe there's a delay while the async idle sending block gets compiled on first run.

I'm going to have to leave this issue alone until Tuesday, very busy till then. Cheers.

JMurph2015:
I was just going to put it out there that I'll just be kinda listening from here until we get back around to how the IJulia-side interface should be structured, just because I don't know enough about Interact.jl and Reactive.jl internals to be useful currently.  If anyone has changes on the code here, just tag me and I'll be along asap.

stevengj:
It sounds to me like we may be able to just ditch the delay entirely in IJulia and not worry about any idle-synchronization methods.

If the only point of #491 was to avoid truncation of STDERR messages from Interact, I think the right solution is for Interact not to use STDERR at all but rather to send an `error` message directly.

JobJob:
Hi, sorry for the delay, busy week.

> If the only point of #491 was to avoid truncation of STDERR messages from Interact, I think the right solution is for Interact not to use STDERR at all but rather to send an error message directly.

It mostly was, but the delay in sending the idle can help output to appear under the right cell, in some circumstances. It's not particularly elegant or reliable though. Something like what we have in this PR would be good, in general, since it would make it possible to have output from long running processes appear under the correct cell. You would need to define e.g.:
```
function cellprintln(exec_msg, args...; name="STDOUT")
    send_ipython(publish[],
             msg_pub(exec_msg, "stream",
                     Dict("name" => name, "text" => string(args...))))
end
```
and then in a cell you could have
```
IJulia.busy()
cell_exec_msg = IJulia.execute_msg
@async begin
    sleep(10.0) # simulation of long running process
    cellprintln(cell_exec_msg, "important information")
    IJulia.idle()
end
```

If you then run other cells while that cell is doing its thing, the output ("important information") would correctly appear under the cell above.

Without holding the idle, the output wouldn't appear with the `cellprintln` above (since the idle disables output addressed by the cell's most recent execute_msg), though it would appear with a regular println, only it would be below the most recently executed cell. So if you ran other cells in between, it would appear below them.

This won't fix the Interact issue directly, but I'm pretty sure the solution I mentioned in the last paragraph of [this comment above](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-315364663) will fix that problem more neatly, and more generally - not just for errors that are caught by Reactive.

So, in short, I recommend we merge a simple form of this PR, like that proposed above in [this comment](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-314107726) and remove the `idle_delay[]` completely.

stevengj:
I'm still not sure I see the point, since calling `busy`/`idle` is effectively equivalent to just blocking until the task is done:
```jl
t = @async begin
    sleep(10.0) # simulation of long running process
    println("important information")
end
# do other stuff here
wait(t)
```

JobJob:
Because you can run other cells in the meantime, assuming you put the waitidle in an async block at the end of the eventloop as per the idle_delay currently.

stevengj:
Does Jupyter allow you to execute other cells if a given cell hasn't sent an idle message yet?  If so, then I'm fine with the simplified version of this PR as suggested above.

JobJob:
Yes, try setting the current idle delay to 10.0, you'll still get the results of multiple cells.

stevengj:
Sounds good then; a lock is definitely better than an idle delay.

JMurph2015:
Hi all,
So I was reading through and it sounds like the solution from [this comment](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-315364663) is pretty popular, is that to say that it should be exactly that implementation or are my most recent commits sufficiently far down that road?

stevengj:
Do you mean the solution from [this comment](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-314107726), i.e. the `busy`/`idle`/`waitidle()` functions I proposed?  That's what I would prefer.

JMurph2015:
Oops, yeah, copied the wrong link there.  The one problem I have with that is that there is no quick way to implement that on a package.  You need to either build it in from the start or track all of your asynchronicity in from the start.

stevengj:
Why is it hard for a package to add `busy` and `idle` calls?

JMurph2015:
I think that knowing that something should have places to put locks is different than actually finding the correct locations to make sure it catches all of the asynchronous parts.  If you are writing something from scratch its easy to plan accordingly.  But similar to the case of Interact, an existing project could be challenging to refactor to enable that sort of flow.

JobJob:
@JMurph2015 as I understand it, this PR currently enables the type of workaround mentioned in your comment above: https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-314825206

Just to rehash that comment, a package could implement, say, a 0.5 sec delay before sending the idle on every request, by running something like this:
```
@async begin
    while true
        IJulia.wait_for_ready()
        IJulia.busy()
        sleep(0.5)
        IJulia.idle()
    end
end
```

It's implemented with two separate level triggers:
`idle_hold_cond` controlled by `busy()` and `idle()`, which IJulia waits on before sending the idle (as per stevengj's simple version) and an additional
`idle_ready_event` which a package can wait on to do something.

I mean, maybe no one's ever going to use it, but I suppose it doesn't hurt. Maybe we just add the above example to the docstring of `wait_for_ready()`?

Also you get rid of all the Interact specific stuff, I'll sort Interact out once something is merged here.

src/eventloop.jl
```diff
@@ -18,9 +18,14 @@ function eventloop(socket)
                 end
             finally
                 @async begin
-                    sleep(idle_delay[])
+                    idle_ready_event += 1
```
JobJob: Shouldn't the `idle_ready_event` be initialised to 1, and this be `-= 1` and the other one below increment it? This needs to notify to anyone waiting on `idle_ready_event` no?

src/eventloop.jl
```diff
@@ -52,7 +57,40 @@ function waitloop()
...
+    Wait until IJulia is ready to send an "status":"idle" message.
+"""
+function wait_for_ready()
+    wait(idle_ready_event)
+end
+
+# Define an even more simple level triggered event so that we can easily wait on it.
+import Base.+, Base.-, Base.notify, Base.wait
+mutable struct LevelTrigger
+    state::Ref{Int}
```
JobJob: This could just be an Int, no? Unless you want to make the LevelTrigger a regular (immutable) struct. We just use Ref's at the module level to be able to have const globals, for performance reasons (non-const globals are slow). Maybe I'm missing something?

stevengj:
Honestly, I'm very reluctant to merge this at all, since the need for it seems purely hypothetical, and I definitely don't want to merge a complicated PR that implements its own `LevelTrigger` type unless there is a clear need.  The only packages that would need this would be IJulia-specific, and such packages seem mostly likely to be better off sending `display` or `error` messages directly.  And it's hard to reason about what functionality might be necessary here without a concrete use-case.

I'd rather just fix Interact to send `error` messages (rather than `STDERR`) and eliminate the idle delay.

src/eventloop.jl
JMurph2015: Oh, yeah, I had totally missed that I had a non-const global in there. It has to be mutable for the struct to be useful, but we could probably just do the integer increment-decrement solution.  I think the complexity-benefit there is better.

JMurph2015:
If there weren't a delay at all, I wouldn't mind, but I don't I'm not involved with any packages taking advantage of these sort of features.  I can probably clean up the code some more to make it simpler, but the only reason I didn't use any of the dozen or so packages that export a level-triggered type was to avoid adding a dependency.

stevengj:
We don't need a level triggered type; see my proposed simpler code. 

I suggest that we fix Interact, then remove the delay, then return to the possible feature discussed in this PR if a real need arises.

JMurph2015:
Not asking to merge right now, but I thought it was useful to try to show in code what I meant about there being a use case I've been saying was missing from @stevengj 's solution.  Also this way, if this PR ever gets resurrected, it'll have this already done.

stevengj:
The use case for `wait_for_ready()` seems even more hypothetical than `busy()`.

stevengj:
The problem with implementing features based on hypothetical needs is twofold.  First, you may be unnecessarily complicating the code.  Second, even if a real version of the need arises, it may be different from what you guessed and may require different semantics.

JMurph2015:
I think the use case for `wait_for_ready()` is pretty un-hypothetical, it would be used to immediately fix Interact's bugs until better analysis was done on code paths that could lead to clipped output.  But beyond that, yes I do agree that it has "pretty niche" use cases, none of which are super obvious to me.  To me it just makes sense considering we are exposing a pseudo-async interaction with IJulia already, so might as well allow it to go the other way.

stevengj:
As I explained, this is the wrong way to fix Interact.  A quick fix is pretty easy: don't use STDERR for error messages.

JMurph2015:
So then where does that leave us?  Should we just start another PR to remove the idle_delay[]?  Do we make no changes to IJulia?  Or do we go forward with some further simplified version?

stevengj:
As I said above, my feeling is that we should leave IJulia as-is, fix Interact, remove the delay from IJulia, and return to this PR if and when a real need arises.

JobJob:
> Don't use STDERR for error messages

Not sure if it was clear from the above - https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-315391648 - but it seems the display of a Jupyter "error" message has the same problems as displaying stream messages. If you send a stream message instead of an error message in your code in https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-315377812, I'm pretty sure the results are the same.

You may get some advantage in that you send the message synchronously, but it is not generally reliable for getting output to appear below a particular cell - if the idle has been sent, it won't display.

As I mentioned [above (last paragraph)](https://github.com/JuliaLang/IJulia.jl/pull/563#issuecomment-315364663), the most general workaround for Interact's problem is to not send the comm message that interferes with displaying output below the last executed cell. I think it's probably avoidable.

---

In summary, if we want to enable the possibility of addressing output to a particular cell, we need something like this PR to delay sending the idle. 

If not, and we're happy with the status quo of output usually displaying below the last executed cell (which mirrors IPython), then we can do as Steven has said in the comment immediately above.

Unfortunately I may not have much time to try this alternate workaround for Interact for the next month or so. If people are desperate to remove the idle delay asap, I'd be fine with having the default be 0.0, and having Interact set it to 0.5.

stevengj:
@JobJob, I can't reproduce your problem.  Even if I set `IJulia.idle_delay[] = 0`, and `sleep` for 10 seconds, and execute several other cells in the meantime, my example code from above where I send an `"error"` message still works (the error output goes to the correct cell).

JobJob:
That's interesting, I still can. What Jupyter you using? I'm on 4.2.3

stevengj:
I have Jupyter 4.3.0

stevengj:
Sending a `stream` message also works for me.  Maybe they fixed this?

JobJob:
You're running the exact code from your comment above? Just with setting the idle_delay before that?

> Sending a stream message also works for me. Maybe they fixed this?

Nice, was just going to ask. Mayyybeee...

Anyway, I'm outta here for now. Very interesting development 😄

stevengj:
Yes.  The exact code with the sleep increased to `sleep(10)`, first setting `IJulia.idle_delay[] = 0`.

JMurph2015:
Also, does this mean we can get rid of the delay by default?  I'll open another PR for that if so.

stevengj:
More than just setting the default delay to zero, I would get rid of the delay code entirely once Interact is fixed, since it was only added for Interact.  If the need ever arises again, we can return to a version of this PR.

PR #569, opened 2017-07-08
Get the latest syntax highlighter at build time

pabloferz:
As proposed [here](https://discourse.julialang.org/t/rfh-updating-highlighters-for-0-6-syntax/4650), this would try to download the latest version of the syntax highlighter from CodeMirror when the miniconda installation via was followed.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -40,6 +40,26 @@ end
 info("Found Jupyter version $jupyter_vers: $jupyter")
 
 #######################################################################
+# Get the latest syntax highlighter file.
+if isconda
+    highlighter = joinpath(Conda.LIBDIR, "python2.7", "site-packages", "notebook", "static",
+                           "components", "codemirror", "mode", "julia", "julia.js")
+    # CodeMirror commit from which we get the syntax highlighter
+    cm_commit = "d444cd1007bd32e418ace8ffed1cb0e692ee147d"
+    highlighter_url = "https://raw.githubusercontent.com/codemirror/CodeMirror/" *
+                      cm_commit * "/mode/julia/julia.js"
+    if isfile(highlighter)
+        try
+            download(highlighter_url, highlighter)
+        catch e
+            warn("The following error occurred while attempting to download latest ",
+                 "syntax highlighting definitions:\n\n", e, "\n\nSyntax highlighting may ",
+                 "not work as expected.")
+        end
+    end
+end
+
+#######################################################################
 # Warn people upgrading from older IJulia versions:
 try
     juliaprof = chomp(readstring(pipeline(`$ipython locate profile julia`,


pabloferz:
@yuyichao if no Conda installation has been performed then `isfile(highlighter)` will return `false` and no download will occur.  Would that be enough or are you thinking of a more general scenario?

yuyichao:
`isconda`

tkelman:
What's the expectation of master continuing to work properly? Making the code so installation-time-dependent and unpredictable in terms of what you'll get doesn't seem great to me. How often does this get updated?

pabloferz:
@tkelman It's not updated that frequently (at least not the julia mode). Almost all updates in the last one and a half year or so have been made by myself and just one by the main CodeMirror developer. But now that you mentioned it, I agree this solution is suboptimal.

Another option would be to distribute a systax file ourselves that we can maintain and try as much as possible to keep it in sync with the CodeMirror one.

deps/build.jl
```diff
@@ -40,6 +40,15 @@ end
 info("Found Jupyter version $jupyter_vers: $jupyter")
 
 #######################################################################
+# Get the latest syntax highlighter file.
+if isconda
+    highlighter = joinpath(Conda.LIBDIR, "python2.7", "site-packages", "notebook", "static",
+                           "components", "codemirror", "mode", "julia", "julia.js")
+    highlighter_url = "https://raw.githubusercontent.com/codemirror/CodeMirror/master/mode/julia/julia.js"
```
ararslan: If the stability of master is a concern, we could just use a specific commit and bump the commit we use here every now and then.

pabloferz: Good idea. Updated accordingly.

aviks:
Downloading something at build time seems like a very risky operation to me, in that there are so many ways this can break. I would very much prefer to distribute the file directly with IJulia. Even if we are out of sync with CodeMirror for some time, the possibility of breakage seems much lower. 

Also, there are many installations that occur with limited internet access. So users might have a private conda repository but no access to github. So if we do go with this solution, `Pkg.build("IJulia")` must not fail if the download fails.

pabloferz:
@aviks this won't download anything if you are using your own jupyter installation. This will only try to download the file if you're using `Conda` for managing the jupyter installation, case in which you'd need internet access anyway.

pabloferz:
Given some of @aviks's concerns, and as an extra precaution, I've added a `try` block around the download in case it fails so the rest of the installation isn't affected by this. But I reiterate, this codepath won't be touched for jupyter installations other than the ones managed by `Conda`.

deps/build.jl
```diff
@@ -40,6 +40,20 @@ end
...
+# Get the latest syntax highlighter file.
+if isconda
+    highlighter = joinpath(Conda.LIBDIR, "python2.7", "site-packages", "notebook", "static",
+                           "components", "codemirror", "mode", "julia", "julia.js")
+    # CodeMirror commit from which we get the syntax highlighter
+    cm_commit = "d444cd1007bd32e418ace8ffed1cb0e692ee147d"
+    highlighter_url = "https://raw.githubusercontent.com/codemirror/CodeMirror/" *
+                      cm_commit * "/mode/julia/julia.js"
+    try
+        isfile(highlighter) && download(highlighter_url, highlighter)
```
ararslan: Not really necessary, but if you wanted to get really fancy, you could do something like
```julia
if isfile(highlighter)
    try
        download(highlighter_url, highlighter)
    catch
        info("Unable to download latest syntax highlighting definitions. ",
             "Your definitions may be out of date.")
    end
end
```

aviks:
I've seem firewall configurations where the conda servers are whitelisted, while github is not.

deps/build.jl
pabloferz: How about

```julia
info("Unable to download latest syntax highlighting definitions. ",
       "Syntax highlighting might not be rendered properly.")
```

ararslan: Sounds good to me.

pabloferz:
I believe all concerns are now addressed. This should be GTG. I'll merge in a day if no one objects or beat me to it.

deps/build.jl
```diff
@@ -40,6 +40,25 @@ end
...
+                           "components", "codemirror", "mode", "julia", "julia.js")
+    # CodeMirror commit from which we get the syntax highlighter
+    cm_commit = "d444cd1007bd32e418ace8ffed1cb0e692ee147d"
+    highlighter_url = "https://raw.githubusercontent.com/codemirror/CodeMirror/" *
+                      cm_commit * "/mode/julia/julia.js"
+    if isfile(highlighter)
+        try
+            download(highlighter_url, highlighter)
+        catch
+            info("Unable to download latest syntax highlighting definitions. ",
```
stevengj: Probably `warn`, not `info`?

stevengj: Should probably also include the exception in the warning.

deps/build.jl
```diff
@@ -40,6 +40,25 @@ end
...
+    # CodeMirror commit from which we get the syntax highlighter
+    cm_commit = "d444cd1007bd32e418ace8ffed1cb0e692ee147d"
+    highlighter_url = "https://raw.githubusercontent.com/codemirror/CodeMirror/" *
+                      cm_commit * "/mode/julia/julia.js"
+    if isfile(highlighter)
+        try
+            download(highlighter_url, highlighter)
+        catch e
+            warn("Unable to download latest syntax highlighting definitions. ",
+                 "Syntax highlighting may not work as expected.\n\n", e)
```
ararslan: Might be clearer to do something like this:
```julia
warn("The following error occurred while attempting to download the latest ",
     "syntax highlighting definitions:\n", e, "\nSyntax highlighting may not ",
     "work as expected.")
```

malmaud:
I am going to advocate @aviks  approach - we ship the syntax file as part of IJulia and install it at the same time as we install the kernel. Since the syntax of Julia is still being updated regularly, that seems like the surest path to making sure everyone has working syntax highlighting.

stevengj:
I would think that any installation using Conda.jl would have internet access.  I would rather not commit the syntax file into IJulia's git repository, preferring to keep the CodeMirror repo as the canonical repo for this.

malmaud:
I'm just unclear what's special about Conda.jl - why wouldn't people not using Conda.jl not want the updated syntax file?

stevengj:
@malmaud, what is special about Conda.jl is that we have write permission to `Conda.LIBDIR`.

pabloferz:
I haven't figured a way to tell jupyter in general to use another syntax file for julia (it might be possible through some `custom.js` file in `~/.jupyter`, but I'd have to look into it deeper). As @stevengj pointed out, we are sure we can overwrite the jupyter installation file only when `Conda.jl` is used.

For the sake of pushing this forward, I'd suggest we merge this and open an issue to remind us to look for a more general solution.

stevengj:
@minrk, @Carreau, is there a way to write something to `~/.jupyter` to tell Jupyter to use a custom syntax file for Julia?

malmaud:
Ah right, good point about permissions. What about if IJulia just defines a function like `IJulia.update_syntax_file()` that users can call themselves if they explicitly want to install the syntax file using whatever python their `PyCall` is using? 

But I have no objection to merging this as-is and worrying about that later.

Carreau:
> @minrk, @Carreau, is there a way to write something to ~/.jupyter to tell Jupyter to use a custom syntax file for Julia?

If you adapt your `julia.js` as follow : 

```
define(function() {

function onload(){

  CodeMirror.defineMode("julia", function(config, parserConf) {
  ... rest of file...
  CodeMirror.defineMIME("text/x-julia", "julia");
}

return {onload:onload}

});
```

And rename it `kernel.js` that you drop next to `kernel.json` in your kernelspec, it should overwrite the default codemirror mode. The drawback (as you monkeypatch JS) is that the notebook will autoreload if you switch _away_ from the current kernel. 

Does that solves your issue ?

stevengj:
@Carreau, does that end up loading both the old codemirror mode and the new one?

Carreau:
It should load both, but only the kernel.js one should be applied. At least
locally for me I get only the kernel.js one. It took me a couple of browser
refresh to have it picked up though.

PR #598, opened 2017-11-12
Fix findfirst depreciation

### LINKED ISSUE #597: Depreciation syntax Julia 0.7 issue
In Julia 0.7,
`findfirst(A, v) is deprecated, use findfirst(equalto(v), A) instead.`
However, this does not work for
execute_request.jl line 53
```
function undisplay(x)
    i = findfirst(displayqueue, x)
    if i > 0
        splice!(displayqueue, i)
    end
    return x
end
```
I suspect `equalto` does not handle `typeof(x)` well.

Nosferican:
`equalto` is only available in Julia 0.7. When support for Julia 0.6 is dropped, it can be changed to use `equalto`.

Closes #597.

### DIFF
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -264,7 +264,7 @@ push_postexecute_hook(f::Function) = push!(postexecute_hooks, f)
 Remove a function `f()` from the list of functions to
 execute after executing any notebook cell.
 """
-pop_postexecute_hook(f::Function) = splice!(postexecute_hooks, findfirst(x -> x == postexecute_hooks, f))
+pop_postexecute_hook(f::Function) = @compat splice!(postexecute_hooks, findfirst(equalto(f), postexecute_hooks))
 
 const preexecute_hooks = Function[]
 """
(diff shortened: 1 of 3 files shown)


stevengj:
`equalto` is provided in Julia 0.6 [by the Compat package](https://github.com/JuliaLang/Compat.jl/commit/3dbca5dd7ca2430ead5cb2c22e1701eeecef1128), as long as we require Compat 0.33.0 or later.

Nosferican:
I bumped Compat to 0.33.0. Is it okay to keep that syntax if we require Julia 0.5?

src/IJulia.jl
```diff
@@ -264,7 +264,7 @@ push_postexecute_hook(f::Function) = push!(postexecute_hooks, f)
 Remove a function `f()` from the list of functions to
 execute after executing any notebook cell.
 """
-pop_postexecute_hook(f::Function) = splice!(postexecute_hooks, findfirst(x -> x == postexecute_hooks, f))
+pop_postexecute_hook(f::Function) = @compat splice!(postexecute_hooks, findfirst(equalto(f), postexecute_hooks))
```
stevengj: You don't need the `@compat` macro just for a function like `equalto` — this function is provided normally by Compat.  `@compat` is only required to rewrite syntax.

Nosferican: Cool, I removed `@compat` then.

stevengj:
Thanks!

PR #602, opened 2017-12-14
fixed lowercase move to Unicode

merl-dev:


stevengj:
This needs to wait for Compat support, so that it can do `using Compat.Unicode` and still work in older Julia versions.

PR #606, opened 2018-01-01
Fix to allow installation on FreeBSD 11.1 STABLE

RobBlackwell:
FreeBSD appends Python version numbers to jupyter commands which breaks the IJulia build process. This quick and dirty fix allows installation for Python 3.6 on FreeBSD (the most recent version in FreeBSD 11.1 STABLE).

Are there any alternative workarounds using environment variables? Perhaps there could be a jupyterkernelspec environment variable for this?

### DIFF
--- a/deps/kspec.jl
+++ b/deps/kspec.jl
@@ -75,7 +75,11 @@ function installkernel(name::AbstractString, julia_options::AbstractString...;
             run(`$jupyter kernelspec install --replace --user $juliakspec`)
             push!(kspec_cmd, jupyter, "kernelspec")
         catch
-            @static if Compat.Sys.isunix()
+
+            @static if Compat.Sys.isbsd()
+                run(`$jupyter-kernelspec-3.6 install --replace --user $juliakspec`)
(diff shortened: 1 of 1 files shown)


stevengj:
`jupyter3.6 kernelspec` doesn’t work?

RobBlackwell:
No, unfortunately it seems like the closest thing is:

```
jupyter -kernelspec-3.6
```

stevengj:
That's frustrating; their packaging is broken if the `jupyter` subcommands don't work regardless of the name of the `jupyter` executable.

RobBlackwell:
Perhaps I'm missing something. I have raised the issue on the freebsd-python mailing list.

https://lists.freebsd.org/pipermail/freebsd-python/2018-January/013124.html

stevengj:
Note that you run `jupyter kernelspec ...`, not `jupyter -kernelspec ...`, to run subcommands.   By requiring a version number on the subcommands, they broke the API of the `jupyter` command-line program.

(Changing the name of *only* the `jupyter` executable itself would have been fine.  `jupyter3.6 kernespec ...` is no problem.)

PR #620, opened 2018-01-31
Add %qtconsole magic

mtikekar:


src/init.jl
```diff
@@ -32,11 +32,18 @@ const read_stdout = Ref{Base.PipeEndpoint}()
 const read_stderr = Ref{Base.PipeEndpoint}()
 const socket_locks = Dict{Socket,ReentrantLock}()
 
+connection_file = ""
+
+function qtconsole()
+    spawn(`$jupyter qtconsole --existing $connection_file`)
```
stevengj: Probably should throw an error if `isempty(connection_file)` is `true`

src/init.jl
```diff
@@ -32,11 +32,20 @@ const read_stdout = Ref{Base.PipeEndpoint}()
 const read_stderr = Ref{Base.PipeEndpoint}()
 const socket_locks = Dict{Socket,ReentrantLock}()
 
+function qtconsole()
+    if inited
+        spawn(`$jupyter qtconsole --existing $connection_file`)
+    else
+        warn("IJulia is not running. qtconsole must be called from an IJulia session.")
```
stevengj: Should be an error, not a warning.

PR #621, opened 2018-01-31
Embed vega-lite JSON directly in JSON message

davidanthoff:
I find this a bit weird, but both Jupyterlab and nteract expect it this way. For MIME types that are really JSON, they want the JSON of the payload to be just part of the message payload, not inside a string. That is how they then also embed it in the notebook files.

I've tested things with this PR, and it all works end-to-end on Jupyterlab and nteract with VegaLite.jl.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -25,9 +25,9 @@ metadata(x) = Dict()
 # return a String=>String dictionary of mimetype=>data
 # for passing to Jupyter display_data and execute_result messages.
 function display_dict(x)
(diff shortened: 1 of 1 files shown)


davidanthoff:
Great! Could you also tag a version?

stevengj:
Note that at some point we probably want to implement a version of this that avoids re-parsing.

The best thing would be to add something to the JSON.jl package, a `JSONString` type that is printed as-is in the JSON output.

davidanthoff:
Yes, that sounds right.

PR #622, opened 2018-01-31
Allow failures for builds on julia nightlies

davidanthoff:
This will turn all the badges green.


stevengj:
Okay.  At some point we will have to fix it on 0.7, though. 😉

davidanthoff:
When they have an alpha? Like six weeks ago ;)

PR #630, opened 2018-02-22
Add support for the dataresource MIME type

davidanthoff:
This one is support out-of-the-box by nteract, and I believe Jupyterlab might get it soon too.


stevengj:
What is this MIME type?

src/execute_request.jl
```diff
@@ -29,6 +30,9 @@ function display_dict(x)
     if mimewritable(application_vnd_vegalite_v2, x)
         data[string(application_vnd_vegalite_v2)] = JSON.parse(limitstringmime(application_vnd_vegalite_v2, x))
     end
+    if mimewritable(application_vnd_dataresource, x)
+        data[string(application_vnd_dataresource)] = JSON.parse(limitstringmime(application_vnd_dataresource, x))
```
stevengj: Please use `JSONText(limitstringmime(application_vnd_dataresource, x))` (using https://github.com/JuliaIO/JSON.jl/pull/237), and similarly for application_vnd_vegalite_v2.

Also update the REQUIRE file to list `JSON 0.17.0`

davidanthoff:
This is a MIME type for tabular data. nteract (and I think jupyterlab eventually out of the box, currently I think with an extension) then use a fancy HTML interactive table UI to show this.

JSON v0.17.0 is not yet registered because the tests fail, so probably best to wait until that is resolved and we can run the test properly here.

src/execute_request.jl
davidanthoff: Done.

davidanthoff:
I think this is ready to be merged now.

PR #632, opened 2018-02-27
Add metadata to complete_reply

mpastell:
This adds metadata to `complete_reply` to be displayed by JupyterLab. The implementation could propably be smarter and I'd happy to revise this based on comments. See: https://github.com/jupyterlab/jupyterlab/issues/3705 for info about the metadata.

jupyter_completer

### DIFF
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -107,7 +107,6 @@ function complete_types(comps)
             if typeof(expr) == Symbol
                 try
                     ctype = complete_type(eval(Main, :(typeof($expr))))
-                catch
                 end
             elseif !isa(expr, Expr)
                 ctype = complete_type(expr)
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -49,6 +49,34 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
     VersionNumber(m.header["version"]) ≥ v"5.2" ? chr2ind(str, ic) : utf16_to_ind(str, ic)
 Base.ind2chr(m::Msg, str::String, i::Integer) = i == 0 ? 0 :
     VersionNumber(m.header["version"]) ≥ v"5.2" ? ind2chr(str, i) : ind_to_utf16(str, i)
+#Compact display of types for Jupyterlab completion
+complete_type(T) = string(T) # maybe shorten?
+complete_type(::Type{<:Function}) = "function"
+complete_type(::Type{<:Type}) = "type"
+complete_type(::Type{<:Tuple}) = "tuple"
+
+function complete_type(T::DataType)
+    s = string(T)
+    (Compat.textwidth(s) ≤ 20 || isempty(T.parameters)) && return s
+    buf = IOBuffer()
+    print(buf, T.name)
+    position(buf) > 19 && return String(take!(buf))
+    print(buf, '{')
+    comma = false
+    for p in T.parameters
+        s = string(p)
+        if position(buf) + sizeof(s) > 20
+            comma || print(buf, '…')
+            break
+        end
+        comma && print(buf, ',')
+        comma = true
+        print(buf, s)
+    end
+    print(buf, '}')
+    return String(take!(buf))
+end
+
 
 function complete_request(socket, msg)
     code = msg.content["code"]
@@ -66,10 +94,8 @@ function complete_request(socket, msg)
 
     codestart = find_parsestart(code, cursorpos)
     comps, positions = Base.REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
-    metadata = Dict()
-
-
     positions += codestart-1
+    metadata = Dict()
     if isempty(comps)
         # issue #530: REPLCompletions returns inconsistent results
         # for positions when no completions are found
@@ -84,11 +110,7 @@ function complete_request(socket, msg)
         for c in comps
             try
                 t = eval(Main, parse("typeof($c)"))
-                if issubtype(t, Function)
-                    push!(typeMap, Dict("text" => c, "type" => "function"))
-                else
-                    push!(typeMap, Dict("text" => c, "type" => "$t"))
-                end
+                push!(typeMap, Dict("text" => c, "type" => "$(complete_type(t))"))
             catch E
             end
         end
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -57,6 +57,7 @@ function complete_request(socket, msg)
     if all(isspace, code[1:cursorpos])
         send_ipython(requests[], msg_reply(msg, "complete_reply",
                                  Dict("status" => "ok",
+                                              "metadata" => Dict(),
                                               "matches" => String[],
                                               "cursor_start" => cursor_chr,
                                               "cursor_end" => cursor_chr)))
@@ -65,6 +66,9 @@ function complete_request(socket, msg)
 
     codestart = find_parsestart(code, cursorpos)
     comps, positions = Base.REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
+    metadata = Dict()
+
+
     positions += codestart-1
     if isempty(comps)
         # issue #530: REPLCompletions returns inconsistent results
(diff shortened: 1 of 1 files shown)


stevengj:
I think you mean https://github.com/jupyterlab/jupyterlab/pull/3508

mpastell:
@stevengj , I hadn't seen that issue, but I'll look into it. The one I linked to is also relevant and links to jupyterlab implementation..

stevengj:
Do we want to combine any other types besides `Function`?   I'm thinking of a `complete_type(T)` function along the lines:

```jl
complete_type(T) = string(T) # maybe shorten?
complete_type(::Type{<:Function}) = "function"
complete_type(::Type{<:Type}) = "type"
complete_type(::Type{<:Tuple}) = "tuple"
```
As noted by the comment above, we should probably also shorten the type if there are two many parameters.  For example:
```jl
julia> typeof(view([1,2,3], 1:2))
SubArray{Int64,1,Array{Int64,1},Tuple{UnitRange{Int64}},true}
```
is probably longer than useful... would be good to display this as `SubArray{Int64,1}`.  We can certainly special-case `SubArray`, but there are many types like this and it would be good to do it more generally.

Maybe something like
```
function complete_type(T::DataType)
    s = stripmodules(string(T))
    (textwidth(s) ≤ 20 || isempty(T.parameters)) && return s
    buf = IOBuffer()
    print(buf, T.name)
    position(buf) > 19 && return String(take!(buf))
    print(buf, '{')
    comma = false
    for p in T.parameters
        s = string(p)
        comma && print(buf, ',')
        comma=true
        if position(buf) + sizeof(s) > 19
            print(buf, "…}")
            return String(take!(buf))
        end
        print(buf, s)
    end
    printf(buf, '}')
    return String(take!(buf))
end
```
This prints the abovementioned `SubArray` type as `"SubArray{Int64,1,…}"`, for example.

It might also be good to strip out the module name from the type, if it is shown, e.g. show `Foo` not `MyModule.Foo`.

mpastell:
I modified this to include your `complete_type` definitions and the output if definitely improved. I couldn't come up with any cases that would be displayed badly and I'll do some more testing later. 

I agree that it would be nice to strip out the module name. 

Jupyterlab also lowercases the type names which does not look so good with Julia type names.

src/handlers.jl
```diff
@@ -84,11 +110,7 @@ function complete_request(socket, msg)
         for c in comps
             try
                 t = eval(Main, parse("typeof($c)"))
-                if issubtype(t, Function)
-                    push!(typeMap, Dict("text" => c, "type" => "function"))
-                else
-                    push!(typeMap, Dict("text" => c, "type" => "$t"))
-                end
+                push!(typeMap, Dict("text" => c, "type" => "$(complete_type(t))"))
```
stevengj: No need for interpolation since `complete_type` already returns a string.  Just use `"type" => complete_type(t)`

src/handlers.jl
```diff
@@ -75,10 +105,21 @@ function complete_request(socket, msg)
     else
         cursor_start = ind2chr(msg, code, prevind(code, first(positions)))
         cursor_end = ind2chr(msg, code, last(positions))
+
+        typeMap = []
+        for c in comps
+            try
+                t = eval(Main, parse("typeof($c)"))
+                push!(typeMap, Dict("text" => c, "type" => "$(complete_type(t))"))
+            catch E
```
stevengj: Just `catch` is fine since you don't use `E`

stevengj:
What does Jupyter display if some of the completions have types and the others do not (because the `try` block fails, e.g. if the symbol isn't defined)?

src/handlers.jl
```diff
@@ -75,10 +105,21 @@ function complete_request(socket, msg)
     else
         cursor_start = ind2chr(msg, code, prevind(code, first(positions)))
         cursor_end = ind2chr(msg, code, last(positions))
+
+        typeMap = []
+        for c in comps
+            try
+                t = eval(Main, parse("typeof($c)"))
```
stevengj: `Meta.parse` to work for Julia 0.7

stevengj:
I'm concerned that just evaluating the completion, assuming it is a valid symbol, could give unexpected results.

For example, what if you are tab-completing a filename in a string?

Or what if you are completing `Base._p`, which gives completions like `"_prod"` that are only defined if you evaluate them in the Base module?

stevengj:
Worse, what if the completion returns a function call?   i.e. what if the replcompletions package somehow returns the string `rm(".", recursive=true, force=true)`?    `eval(parse(...))` seems very unsafe.

mpastell:
Very good points! Do you think the latest implementation is safe? It calls eval only when `typeof(parse(c)) == Symbol` and checks the expression head to display type for macros.

Jupyter will still display the completions for everything else as well, just without type information. The same applies to:

> What does Jupyter display if some of the completions have types and the others do not (because the try block fails, e.g. if the symbol isn't defined)?

src/handlers.jl
```diff
@@ -49,6 +49,51 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+    print(buf, '}')
+    return String(take!(buf))
+end
+
+#Get typeMap for Jupyter completions
+function complete_types(comps)
+    typeMap = []
+    for c in comps
+        expr = Meta.parse(c, raise=false)
+        if typeof(expr) == Symbol
```
stevengj: `!isa(expr, Expr)` should be sufficient here.  That way it will also allow literal numbers and strings.

src/handlers.jl
```diff
@@ -49,6 +49,51 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+            try
+                t = eval(Main, Meta.parse("typeof($c)"))
```
stevengj: No need to parse a second time.  Do `eval(Main, :(typeof($c)))`.

src/handlers.jl
```diff
@@ -49,6 +49,51 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+                push!(typeMap, Dict("text" => c, "type" => "$(complete_type(t))"))
```
stevengj: Just `"type" => complete_type(t)`, since `complete_type` already returns a string.

src/handlers.jl
```diff
@@ -49,6 +49,51 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+            catch
+            end
+        elseif expr.head == :macrocall
```
stevengj: Note that this assumes `isa(expr, Expr)`, but this is now correct with my suggested change above.

src/handlers.jl
```diff
@@ -49,6 +49,51 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+    end
+    print(buf, '}')
+    return String(take!(buf))
+end
+
+#Get typeMap for Jupyter completions
+function complete_types(comps)
+    typeMap = []
+    for c in comps
+        expr = Meta.parse(c, raise=false)
```
stevengj: Before parsing, we might want to first check if `c` is a keyword.  Do something like:

```jl
if !isempty(searchsorted(sorted_keywords, c))
    push!(typeMap, Dict("text" => c, "type" => "keyword"))
end
```

where `sorted_keywords` is defined above as:
```jl
if isdefined(REPLCompletions, :sorted_keywords)
    import REPLCompletions: sorted_keywords
else
    const sorted_keywords = [
        "abstract type", "baremodule", "begin", "break", "catch", "ccall",
        "const", "continue", "do", "else", "elseif", "end", "export", "false",
        "finally", "for", "function", "global", "if", "import",
        "let", "local", "macro", "module", "mutable struct",
        "primitive type", "quote", "return", "struct",
        "true", "try", "using", "while"]
end
```

Probably better to consolidate the `push!(typeMap, ...)` calls into a single line `isempty(ctype) || push!(typeMap, Dict("text" => c, "type" => ctype))` and just have the other code assign the type to a variable `ctype`, defaulting to `""`.

src/handlers.jl
mpastell: I added another branch as it makes sense to call eval on symbols to get their type.

mpastell:
Thanks for the review. I've tried to fix everything. I added a preview for latex symbols and emojis to auto completion as well, it abuses the Jupyterlab feature a bit but I think it's pretty useful.

src/handlers.jl
```diff
@@ -49,6 +49,76 @@ Base.chr2ind(m::Msg, str::String, ic::Integer) = ic == 0 ? 0 :
...
+            ctype = get(emoji_symbols, c, "")
+            isempty(ctype) || (ctype = "emoji: $ctype")
+        elseif startswith(c, "\\")
+            ctype = get(latex_symbols, c, "")
+        else
+            expr = Meta.parse(c, raise=false)
+            if typeof(expr) == Symbol
+                try
+                    ctype = complete_type(eval(Main, :(typeof($expr))))
+                catch
```
stevengj: You can just omit `catch` if you have an empty catch block.

mpastell: OK, I removed it now.

stevengj:
Thanks for keeping at this!

mpastell:
Thanks for the reviews and for IJulia, it's a great tool!

PR #634, opened 2018-03-03
RFC: Improved tab completion for symbols

mpastell:
This implements improved (IMHO) tab completion for latex symbols and emojis. It completes the list of symbols and names starting with that symbol as well. Is this something that others see as a preferable behavior as well? I have only tested it in Jupyterlab with Julia 0.6 so far.

complete_symbols

### DIFF
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -67,6 +67,7 @@ end
 complete_type(::Type{<:Function}) = "function"
 complete_type(::Type{<:Type}) = "type"
 complete_type(::Type{<:Tuple}) = "tuple"
+complete_type(::Void) = "void"
 
 function complete_type(T::DataType)
     s = string(T)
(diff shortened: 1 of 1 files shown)


stevengj:
This is the kind of patch that should go directly into the REPLCompletions stdlib package, I think.

mpastell:
I think the REPL user experience would be quite poor this patch. In IJulia you can choose the completions from a menu where as in the REPL you need to type the rest of the characters.

Also the emojis and some of the latex symbols only show as question marks in the REPL.

mpastell:
True, but that's an issue that exists in the default windows terminal and also in the default Ubuntu terminal (although for less symbols than in windows) and font  so it will affect a lot of users. 

Also if I hit tab in the terminal and see the character as completion I don't probably remember how to I get the completion without seeing the text. e.g. if see 👹 and 👺I will probably not remember which one is `japanese_goblin` and which `japanese_ogre`

stevengj:
If you are using the REPL in a Terminal where some emoji is not supported by the font, then presumably this PR is irrelevant since you won't be defining variables using that symbol, no?

(Unfortunately, it doesn't seem that we have tab completions for 👹 and 👺.  😛 )

mpastell:
I suppose you'd avoid the characters that are not supported by the terminal. Another thing that you'd need to recognize the symbols in the REPL in order to get the completion. So if you type e.g. `\a`+TAB and see symbols instead of strings you need to remember how type it. 

Probably I should just try to find the time to implement this for the REPL and what people think.

PR #642, opened 2018-03-16
Drop Julia v0.5

rdeits:
Per https://discourse.julialang.org/t/errors-for-git-pkg/9351 Julia v0.5 is no longer supported on Linux and package operations will not be possible. This PR just stops testing Julia v0.5 on Travis since those tests will otherwise permanently fail. 

Personally I would recommend just dropping v0.5 from REQUIRE entirely.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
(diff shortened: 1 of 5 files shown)


stevengj:
If we stop testing it we should also drop it from REQUIRE.

rdeits:
Ok, done.

PR #645, opened 2018-04-04
Fix windows check in ipython.jl

jekyllstein:
This change is required to run in julia 0.6 replacing the @windows macro used in julia 0.5 and earlier


stevengj:
To work in 0.7, it should be `Compat.Sys.iswindows()` with a `using Compat` at the top of the file.

stevengj:
See also #623.

PR #647, opened 2018-04-10
Implement autosave!

cstjean:


src/IJulia.jl
```diff
@@ -246,6 +246,29 @@ is `STDOUT`.
...
+# IPython kernel code: https://github.com/ipython/ipykernel/blob/17bd0602bf0d800d018caae3621483c6e652684a/ipykernel/zmqshell.py#L414
+
+"""
+    autosave!(seconds)
+
+Set the autosave interval in the notebook (in seconds).
+The default value is 120, or two minutes. `autosave(0)` will disable autosave.
+This only has an effect when called from the notebook interface.
+"""
+function autosave!(seconds::Integer)
```
stevengj: Why an integer?   Seems like it should be `::Real`

src/IJulia.jl
```diff
@@ -246,6 +246,29 @@ is `STDOUT`.
...
+    milliseconds = 1000 * seconds      # javascript wants milliseconds
```
stevengj: Does it need to be an integer?   If so, add a `round` call…

stevengj:
This really seems like the sort of thing that should be in a the notebook UI, and should not be handled by the kernel at all.  cc @Carreau.

Carreau:
It should be cover in frontends extensions like https://github.com/ipython-contrib/jupyter_contrib_nbextensions

cstjean:
If it's covered by the extensions, then should I just close this PR? I don't mind either way.

src/magics.jl
```diff
@@ -54,6 +54,12 @@ alias_magic_help(magic::AbstractString, args::AbstractString) = md"""
 
     which you can then run with e.g. `bracket("hello world")`."""
 
+autosave_magic_help(magic::AbstractString, args::AbstractString) = md"""
+    The equivalent of `%autosave 60` magic in Julia is `IJulia.autosave!(60)`,
+    to set the notebook autosave frequency, in seconds. Use `0` to disable
+    autosaving.
```
stevengj: This help could point to the extension instead.

PR #667, opened 2018-06-19
Add logging override for 0.7

staticfloat:
0.7 has a new logging infrastructure, which saves the output stream (e.g. `Base.stderr`) within itself.  We need to override that if we want to receive things like `warn()` or `info()`.  This is not the most elegant way this could be done, but it does seem to work (and passes its own test suite on 0.7)

### DIFF
--- a/src/stdio.jl
+++ b/src/stdio.jl
@@ -160,7 +160,7 @@ function send_stream(name::AbstractString)
         n = num_utf8_trailing(d)
         dextra = d[end-(n-1):end]
         resize!(d, length(d) - n)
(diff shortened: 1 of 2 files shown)


staticfloat:
Hmmm, although things seem to be getting captured, they do not seem to be getting communicated to the browser.

staticfloat:
Ah, found it.  `String(d)` was clobbering `d`, so we need to copy it first.

stevengj:
See also #26798.

stevengj:
Closing in favor of #671

Issue #654, opened 2018-04-19
Compat.stderr != STDERR


cstjean:
I've been seeing some weird buffering in IJulia recently (notably with ProgressMeter.jl, which uses `Compat.stderr`), and I suspect it's because `Compat.stderr == STDERR` is false in IJulia on 0.6 (but true at the REPL). Likewise for STDOUT/STDIN.

stevengj:
It seems there are a lot of problems with redirection interacting with code that copies a reference to `stderr`.  See also e.g. JuliaLang/julia#26798.

I wonder if a more general solution would be to make `stdout` and `stderr` in Julia a mutable container along the lines of:
```
mutable type IOContainer
    io::IO
end
const stdout = IOContainer(...real stdout object...)
```
with `redirect_stdout` mutating `stdout.io = ...new IO...`.   That way existing references to `stdout` would be updated.

stevengj:
But in the meantime IJulia probably needs to rework how it redirects I/O.

stevengj:
@cstjean, have you already checked whether this was fixed on IJulia master by #641?

cstjean:
Yes, that's the version I'm on.

stevengj:
Looks like the problem is that #641 hard-coded `redirect_stdout` to directly change `STDOUT`.

cstjean:
Might be unrelated, but this looks wrong:

![screen shot 2018-04-19 at 21 35 30](https://user-images.githubusercontent.com/8622776/39026207-c0773334-4419-11e8-912f-437a21094c91.png)

I don't know how `Compat.stdout` manages to track STDOUT at the REPL inside of `redirect_stdout`, but whatever it's doing is broken in IJulia.

cstjean:
AFAICT, this was fixed by #671 / not relevant on 1.0

Issue #637, opened 2018-03-13
Build issue on nightly - JULIA_HOME not defined


tk3369:
Looks like `JULIA_HOME` has been [renamed](https://docs.julialang.org/en/latest/NEWS/#Deprecated-or-removed-1)  to `JULIA_BINDIR`.  

```
$ cat /Users/tomkwong/.julia/packages/IJulia/tVAv/deps/build.log
Solving environment: ...working... done

# All requested packages already installed.

┌ Warning: `info()` is deprecated, use `@info` instead.
│   caller = top-level scope at build.jl:31
└ @ Core build.jl:31
INFO: Installing Jupyter via the Conda package.
┌ Warning: `convert(::Type{VersionNumber}, v::AbstractString)` is deprecated, use `VersionNumber(v)` instead.
│   caller = (::getfield(Main, Symbol("#prog_version#5")))(::String) at build.jl:17
└ @ Main build.jl:17
┌ Warning: `info()` is deprecated, use `@info` instead.
│   caller = top-level scope at build.jl:39
└ @ Core build.jl:39
INFO: Found Jupyter version 4.4.0: /Users/tomkwong/.julia/packages/Conda/BTNx/deps/usr/bin/jupyter
┌ Warning: `replace(s::AbstractString, pat, f)` is deprecated, use `replace(s, pat => f)` instead.
│   caller = installkernel(::String) at kspec.jl:36
└ @ Main kspec.jl:36
ERROR: LoadError: UndefVarError: JULIA_HOME not defined
Stacktrace:
 [1] uv_error at ./libuv.jl:68 [inlined]
 [2] unlink(::String) at ./file.jl:549
 [3] #rm#9(::Bool, ::Bool, ::Function, ::String) at ./file.jl:151
 [4] #rm at ./<missing>:0 [inlined]
 [5] #installkernel#6(::String, ::Function, ::String) at /Users/tomkwong/.julia/packages/IJulia/tVAv/deps/kspec.jl:117
 [6] installkernel(::String) at /Users/tomkwong/.julia/packages/IJulia/tVAv/deps/kspec.jl:36
 [7] top-level scope at /Users/tomkwong/.julia/packages/IJulia/tVAv/deps/build.jl:75
 [8] include at ./boot.jl:306 [inlined]
 [9] include_relative(::Module, ::String) at ./loading.jl:1072
 [10] include(::Module, ::String) at ./sysimg.jl:29
[...]

PR #671, opened 2018-07-05
fixes for 0.7

### LINKED ISSUE #661: The kernel appears to have died. It will restart automatically Julia V0.7
Kernel dies when running the nightly build in Julia v0.7. I am using ``` "ZMQ"=> v"0.6.0+"``` and ```"Conda"=> v"0.7.1+"```, ```MbedTLS=> v"0.5.8+```. 

Running ```Pkg.build("IJulia")``` is successful and besides the deprecations, it doesn't give me any error. 

The output of the Jupyter window right before the error is 

```Julia

┌ Warning: The function `cfunction` is now written as a macro `@cfunction`.
│   caller = start_heartbeat at heartbeat.jl:18 [inlined]
└ @ Core heartbeat.jl:18
[I 17:35:19.910 NotebookApp] KernelRestarter: restarting kernel (1/5), keep random ports
WARNING:root:kernel cc809c67-d194-447d-9d74-a5a05f21dc83 restarted


```

fredrikekre:
This is  #623 + #667 + two additional fixes.  Closes #661, #654, #637.

This PR works locally on both 0.7 and 0.6 for me, would be great to get IJulia to work properly on 0.7!

### DIFF
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -38,7 +38,18 @@ export notebook, installkernel
 using ZMQ, JSON, Compat
 import Compat.invokelatest
 using Compat.Unicode: uppercase, lowercase
+import Compat.Dates
 using Compat.Dates: now
+import Compat.Random
+
+if VERSION ≥ v"0.7.0-DEV.2338" # julia#24361
+  using Base64: Base64EncodePipe
+end
+if isdefined(Base, :REPL) && !Base.isdeprecated(Base, :REPL)
+  import Base.REPL
+else
+  import REPL
+end
 
 #######################################################################
 # Debugging IJulia
@@ -198,7 +209,7 @@ function clear_history(indices)
 end
 
 # since a range could be huge, intersect it with 1:n first
-clear_history{T<:Integer}(r::AbstractRange{T}) =
+clear_history(r::AbstractRange{T}) where {T<:Integer} =
     invoke(clear_history, Tuple{Any}, intersect(r, 1:n))
 
 function clear_history()
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -113,10 +113,10 @@ function notebook(; dir=homedir(), detached=false)
     inited && error("IJulia is already running")
     if Compat.Sys.isapple() # issue #551 workaround, remove after macOS 10.12.6 release?
         withenv("BROWSER"=>"open") do
-            p = spawn(Cmd(`$notebook_cmd`, detach=true, dir=dir))
+            p = run(Cmd(`$notebook_cmd`, detach=true, dir=dir); wait=false)
         end
     else
-        p = spawn(Cmd(`$notebook_cmd`, detach=true, dir=dir))
+        p = run(Cmd(`$notebook_cmd`, detach=true, dir=dir); wait=false)
     end
     if !detached
         try
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -243,7 +243,7 @@ function history(io::IO, indices::AbstractVector{Int})
 end
 
 history(io::IO, x::Union{Integer,AbstractVector{Int}}...) = history(io, vcat(x...))
-history(x...) = history(STDOUT, x...)
+history(x...) = history(stdout, x...)
 history(io::IO, x...) = throw(MethodError(history, (x...,)))
 history() = history(1:n)
 """
@@ -256,7 +256,7 @@ The optional `indices` argument is one or more indices or collections
 of indices indicating a subset input cells to print.
 
 The optional `io` argument is for specifying an output stream. The default
-is `STDOUT`.
+is `stdout`.
 """
 history
 
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -25,27 +25,27 @@ metadata(x) = Dict()
 # for passing to Jupyter display_data and execute_result messages.
 function display_dict(x)
     data = Dict{String,Any}("text/plain" => limitstringmime(text_plain, x))
-    if mimewritable(application_vnd_vegalite_v2, x)
+    if showable(application_vnd_vegalite_v2, x)
         data[string(application_vnd_vegalite_v2)] = JSON.JSONText(limitstringmime(application_vnd_vegalite_v2, x))
     end
-    if mimewritable(application_vnd_dataresource, x)
+    if showable(application_vnd_dataresource, x)
         data[string(application_vnd_dataresource)] = JSON.JSONText(limitstringmime(application_vnd_dataresource, x))
     end
-    if mimewritable(image_svg, x)
+    if showable(image_svg, x)
         data[string(image_svg)] = limitstringmime(image_svg, x)
     end
-    if mimewritable(image_png, x)
+    if showable(image_png, x)
         data[string(image_png)] = limitstringmime(image_png, x)
-    elseif mimewritable(image_jpeg, x) # don't send jpeg if we have png
+    elseif showable(image_jpeg, x) # don't send jpeg if we have png
         data[string(image_jpeg)] = limitstringmime(image_jpeg, x)
     end
-    if mimewritable(text_markdown, x)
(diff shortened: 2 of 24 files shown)


src/inline.jl
```diff
@@ -1,6 +1,9 @@
 import Base: display, redisplay
 
-immutable InlineDisplay <: Display end
+@static if !isdefined(Base, :AbstractDisplay) # remove when Compat.jl#482 is merged and tagged
```
stevengj: Shouldn't this be:
```jl
if !isdefined(Base, :AbstractDisplay) && !isdefined(Compat, :AbstractDisplay)
  ...
```
(No need for `@static` in global scope unless the expressions have `ccall` or macros...)

stevengj:
Great, thanks for working on this!  Sorry I haven't had time to bang on it.

Do things like `println` now work in both 0.6 and 0.7?

src/inline.jl
stevengj: Actually, this if statement can be deleted completely as the `AbstractDisplay` definition was tagged in Compat v0.52.0

fredrikekre:
Pushed one more deprecation fix and updated to use `Compat.AbstractDisplay` directly.

> Do things like println now work in both 0.6 and 0.7?

Yes! :tada:

src/stdio.jl
```diff
@@ -27,36 +27,51 @@ Base.setup_stdio(io::IJuliaStdio, readable::Bool) = Base.setup_stdio(io.io.io, r
...
+    sq = QuoteNode(Symbol(s))
     @eval function Base.$f(io::IJuliaStdio)
         io[:jupyter_stream] != $s && throw(ArgumentError(string("expecting ", $s, " stream")))
-        eval(Base, Expr(:(=), $S, io))
+        if isdefined(Base, :stdout)
+            # On Julia 0.7, we need to override the global logger as well
+            logger = Base.CoreLogging._global_logstate.logger
+
+            # Override logging if it's pointing at the default stderr
+            if logger.stream == Base.stderr
```
fredrikekre: Actually, there is something bugged here; this is only true on the first call, don't we need to reset it somewhere @staticfloat ? On this branch; with a `@show logger.stream == Base.stderr` inserted:
```
julia> using IJulia

julia> mktemp() do path, io
           redirect_stderr(IJulia.IJuliaStdio(io, "stderr")) do
               @warn("warn")
           end
           flush(io)
           seek(io, 0)
           @show read(io, String)
       end
logger.stream == Base.stderr = true
read(io, String) = "\e[33m\e[1m┌ \e[22m\e[39m\e[33m\e[1mWarning: \e[22m\e[39mwarn\n\e[33m\e[1m└ \e[22m\e[39m\e[90m@ Main REPL[2]:3\e[39m\n"
"\e[33m\e[1m┌ \e[22m\e[39m\e[33m\e[1mWarning: \e[22m\e[39mwarn\n\e[33m\e[1m└ \e[22m\e[39m\e[90m@ Main REPL[2]:3\e[39m\n"

julia> mktemp() do path, io
           redirect_stderr(IJulia.IJuliaStdio(io, "stderr")) do
               @warn("warn")
           end
           flush(io)
           seek(io, 0)
           @show read(io, String)
       end
logger.stream == Base.stderr = false
read(io, String) = ""
```

fredrikekre: Perhaps it does not matter; this seems to be called only once per session anyways and it works in a notebook
warning

fredrikekre: Would it not be simpler to just replace the global logger instead of checking the internal state?

fredrikekre:
I updated the logging redirect, now things seem to work:
working

src/handlers.jl
```diff
@@ -138,7 +138,7 @@ function complete_request(socket, msg)
 
     codestart = find_parsestart(code, cursorpos)
     comps, positions = Base.REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
```
stevengj: Probably need:
```jl
if VERSION < v"0.7.0-DEV.3500" # julia#25544
    import Base: REPLCompletions
else
    import REPL: REPLCompletions
end
```
above so that we can just call this as `REPLCompletions.completions` with no deprecation warning.

fredrikekre: Thanks, there was a similar `if VERSION` check at the top so I added the imports there.

fredrikekre:
While trying the tab-completion for deprecations I got this
```
BoundsError: attempt to access 0-element Array{Int64,1} at index [0]
```
Seems to be a bug here https://github.com/JuliaLang/IJulia.jl/blob/90b0165cf0902f9ca678f968d63e8834119ccc63/src/handlers.jl#L144-L145 where we try to extract the last element although we just made sure it's empty.

stevengj:
Maybe should be `cursor_start = cursor_end = cursor_chr`

src/handlers.jl
```diff
@@ -134,8 +140,8 @@ function complete_request(socket, msg)
     end
 
     codestart = find_parsestart(code, cursorpos)
-    comps, positions = Base.REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
-    positions += codestart-1
+    comps, positions = REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
+    positions = positions .+ codestart-1
```
fredrikekre: Nevermind, it is this fix that is missing a `.` on the `-` and the range turned into an array. Of course last works for an empty range.

src/handlers.jl
```diff
@@ -134,8 +140,8 @@ function complete_request(socket, msg)
     end
 
     codestart = find_parsestart(code, cursorpos)
-    comps, positions = Base.REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
-    positions += codestart-1
+    comps, positions = REPLCompletions.completions(code[codestart:end], cursorpos-codestart+1)
+    positions = positions .+ codestart .- 1
```
fredrikekre: Hmm, still not correct; this returns a `Vector` on Julia 0.6.

stevengj:
LGTM, ready to merge?

src/IJulia.jl
```diff
@@ -198,7 +202,7 @@ function clear_history(indices)
 end
 
 # since a range could be huge, intersect it with 1:n first
-clear_history{T<:Integer}(r::AbstractRange{T}) =
+clear_history(r::AbstractRange{T}) where {T<:Integer} =
```
stevengj: Can just be `clear_history(r::AbstractRange{<:Integer})` since we don't use `T`.

src/execute_request.jl
```diff
@@ -58,30 +56,50 @@ const displayqueue = Any[]
 
 # remove x from the display queue
 function undisplay(x)
-    i = findfirst(equalto(x), displayqueue)
-    if i > 0
+    i = findfirst(isequal(x), displayqueue)
```
stevengj: Shouldn't this be `Compat.findfirst` to get the version that returns `nothing` on 0.6?

src/execute_request.jl
```diff
@@ -58,30 +56,50 @@ const displayqueue = Any[]
...
+    import Base.REPL: ip_matches_func
+else
+    import Base: ip_matches_func
+end
+
 function show_bt(io::IO, top_func::Symbol, t, set)
     # follow PR #17570 code in removing top_func from backtrace
-    eval_ind = findlast(addr->Base.REPL.ip_matches_func(addr, top_func), t)
-    eval_ind != 0 && (t = t[1:eval_ind-1])
+    eval_ind = findlast(addr->ip_matches_func(addr, top_func), t)
```
stevengj: Similar to above, use `Compat.findlast` and then you can skip the `eval_ind != 0` check.

stevengj:
I'm just going to merge this as-is so that people can use IJulia with 0.7, and then add more fixes in another PR.

fredrikekre:
Thanks! I have to admit that I did not review the changes from  #623 myself.

PR #672, opened 2018-07-06
a few more 0.7 updates

stevengj:
continuing #671

### DIFF
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -202,7 +202,7 @@ function clear_history(indices)
 end
 
 # since a range could be huge, intersect it with 1:n first
-clear_history(r::AbstractRange{T}) where {T<:Integer} =
+clear_history(r::AbstractRange{<:Integer}) =
     invoke(clear_history, Tuple{Any}, intersect(r, 1:n))
 
 function clear_history()
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -273,7 +273,8 @@ push_postexecute_hook(f::Function) = push!(postexecute_hooks, f)
 Remove a function `f()` from the list of functions to
 execute after executing any notebook cell.
 """
-pop_postexecute_hook(f::Function) = filter!(isequal(f), postexecute_hooks)
+pop_postexecute_hook(f::Function) =
+    splice!(postexecute_hooks, findlast(isequal(f), postexecute_hooks)
 
 const preexecute_hooks = Function[]
 """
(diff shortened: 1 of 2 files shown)


src/IJulia.jl
```diff
@@ -273,7 +273,8 @@ push_postexecute_hook(f::Function) = push!(postexecute_hooks, f)
 Remove a function `f()` from the list of functions to
 execute after executing any notebook cell.
 """
-pop_postexecute_hook(f::Function) = splice!(postexecute_hooks, findfirst(equalto(f), postexecute_hooks))
+pop_postexecute_hook(f::Function) =
+    splice!(postexecute_hooks, findlast(isequal(f), postexecute_hooks))
```
fredrikekre: `Compat.`?

src/IJulia.jl
```diff
@@ -289,7 +290,8 @@ push_preexecute_hook(f::Function) = push!(preexecute_hooks, f)
 Remove a function `f()` from the list of functions to
 execute before executing any notebook cell.
 """
-pop_preexecute_hook(f::Function) = splice!(preexecute_hooks, findfirst(equalto(f), preexecute_hooks))
+pop_preexecute_hook(f::Function) =
+    splice!(preexecute_hooks, findlast(isequal(f), preexecute_hooks))
```
fredrikekre: `Compat.`?

src/IJulia.jl
```diff
@@ -306,7 +308,8 @@ push_posterror_hook(f::Function) = push!(posterror_hooks, f)
 Remove a function `f()` from the list of functions to
 execute after an error occurs when a notebook cell is evaluated.
 """
-pop_posterror_hook(f::Function) = splice!(posterror_hooks, findfirst(equalto(f), posterror_hooks))
+pop_posterror_hook(f::Function) =
+    splice!(posterror_hooks, findlast(isequal(f), posterror_hooks))
```
fredrikekre: `Compat.`?

src/IJulia.jl
stevengj: Doesn’t need it. If f isn’t found, whether findlast returns 0 or nothing splice! will throw.

stevengj:
Looks like the Travis failure is unrelated

src/IJulia.jl
stevengj: (But I guess it wouldn't hurt to make sure it throws the same exception on 0.6 and 0.7.)

PR #687, opened 2018-07-18
fix #682: using InteractiveUtils on 0.7

### LINKED ISSUE #682: `using Compat.InteractiveUtils` needed somewhere
`Base.@which`, `Base.@code_XXX`, etc. have associated deprecation warnings on 0.7 -- I think the fix is to include `using Compat.InteractiveUtils` somewhere, but naively putting this at the top of `src/IJulia.jl` doesn't seem to stop the warnings.

schmrlng:
Thanks for the suggestion! I can confirm that this eliminates the deprecation warning on `0.7.0-beta2`.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -1,4 +1,5 @@
 import IJulia
+using IJulia.Compat.InteractiveUtils
 
 # workaround #60:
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks!

PR #691, opened 2018-07-28
Monkeypatch Compat.stdio to match Base.

### LINKED ISSUE #690: Kernel "Idle" Messages Arrive Before "Stream" Messages
I see that this has been discussed before, à la #238.

I was wondering what if this inconsistency with the Jupyter protocol will be fixed; the [protocol specification](https://jupyter-client.readthedocs.io/en/stable/messaging.html#request-reply) says that "After processing the request and publishing associated IOPub messages, if any, the kernel publishes a status: idle message. **This idle status message indicates that IOPub messages associated with a given request have all been received.**" (emphasis mine).

I discovered this while writing a kernel frontend and getting no responses to my `println("foo")` execution requests (the frontend in question considers requests to be done whenever the kernel replies with "idle"), and indeed, if you inspect the WebSocket communications on a normal notebook, you can see that the "status" message arrives before the "stream" message.

travigd:
Fixes #690, #654.

### DIFF
--- a/src/stdio.jl
+++ b/src/stdio.jl
@@ -36,6 +36,9 @@ for s in ("stdout", "stderr", "stdin")
         else
             # On Julia 0.6-, the variables are called Base.STDIO, not Base.stdio
             Core.eval(Base, Expr(:(=), $Sq, io))
+            # We also need to change Compat.stdxxx because we use it in various places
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks!

PR #694, opened 2018-07-31
Make ipy_mime variable more readable

mortenpi:
A purely aesthetic code change. I was consulting this for JuliaDocs/Documenter.jl#764 and figured it might be helpful for other too to have this be a bit more readable. It seems it has just been growing over the years.


stevengj:
Thanks!

Issue #693, opened 2018-07-29
Doesn't work on windows + Julia 0.7


Datseris:
Kernel opens but operations hang indefinitely (e.g. doing `2+2` hangs, never returns).

stevengj:
See https://github.com/JuliaLang/IJulia.jl#debugging-ijulia-problems to get more output.

Datseris:
Hi Steven, I did this but it didn't help me...

You see, what happens is that the computation freezes indefinitely: it *does not* return an error. (doesn't return anything)

![image](https://user-images.githubusercontent.com/19669089/43370094-1e00a96e-9379-11e8-9164-dd3ff9397332.png)

This is the picture I see (forever) when I try to run a computation. User `Marco Concetto Rudilosso` reported same behavior.

Datseris:
Pressing the button `interrupt the kernel` also does nothing.

stevengj:
You have to look in the terminal where you launched `jupyter` to see the [debugging info](https://github.com/JuliaLang/IJulia.jl#debugging-ijulia-problems)

Datseris:
I don't understand.

1. I have done what is said in the documentation paragraph you point (changing the source code).
2. I open a new `notebook()`. I try to run `1+1` and it never returns.
3. I press `interrupt the kernel`, in hope of seeing the "errors" mentioned in the documentation paragraph, but I see nothing because the Julia kernel never dies.
4. I go into the REPL that I typed `notebook()` and I press Ctrl+C to terminate process. All I see is
```
julia> using IJulia

julia> notebook()
WARNING: Force throwing a SIGINT
```
The kernel doesn't actually terminate. I go back to the Jupyter page and it says "The kernel appears to have died. It will restart automatically." or whatever.

And that is all. After that, everything freezes forever **including the REPL**. I can't do anything, no matter how many times I press Cntrl+C I still simply get the message `The kernel appears to have died. It will restart automatically.`

stevengj:
I just tried 0.7 on a Windows machine, and I can reproduce your hang.   However, if I turn on debugging output, then it works!  Arghh.

Turning on debugging output gives me a lot of output (as expeced) in Windows if I run `jupyter notebook` directly in the Windows command shell, but gives me no output with `notebook()`, which is also annoying.

[...]

PR #711, opened 2018-08-16
Move flush_all() out from @async (#693)

d4n1elchen:
This pull request fixes (probably not) #693 

I think this problem should be related to #491 and #563. Previously there's an issue about when to send the idle message. At that time, the solution was adding a delay before sending the idle message. I think it's because of the delay, the code, together with flush_all(), was wrapped by @asnyc. That means the flush_all() won't be called right after message handler finish the jobs. I think that will cause unexpected flush for the following messages. So I move the flush_all() out from the @asnyc wrapping and remain everything being the same (also remove @vprintln I just added). Then everything looks fine. It works with IJULIA_DEBUG is either true or false.

And according to #563, the delay is no more needed and has been removed. I think the @async is also not needed. But I'm not sure about this so I keep it in the code.

### DIFF
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -17,13 +17,8 @@ function eventloop(socket)
                     send_ipython(publish[], msg_pub(execute_msg, "error", content))
                 end
             finally
-                # @async no longer implicitly captures local variables (see PR #618)
                 flush_all()
-                let msg = msg
-                    @async begin
-                        send_status("idle", msg)
-                    end
-                end
+                send_status("idle", msg)
             end
         end
     catch e
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -18,9 +18,9 @@ function eventloop(socket)
                 end
             finally
                 # @async no longer implicitly captures local variables (see PR #618)
+                flush_all()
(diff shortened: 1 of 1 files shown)


Datseris:
@d4n1elchen you god you!!!! This is LGTM!!! Works on Julia 0.7 and Windows 10! Not only that, but it also works from running `using IJulia(); notebook()` from within Julia!!!

YEAAAAAAAAAA

d4n1elchen:
Hey, some bad news here. 😢 

After testing for a while, I found that if the streaming is long, it'll still get stuck.

Datseris:
It also doesn't work when using a package that has to compile. :(

stevengj:
Just get rid of the `@async` entirely.

Honestly, I think #693 is some kind of memory bug, perhaps in ZMQ.jl, perhaps in julia itself.  That's why it disappears and re-appears randomly when you add print messages and make other small unrelated changes.  So I doubt that this PR actually fixes the problem.   But it is still a good change to make.

Datseris:
Steven is right. Turns out this simply "doesn't work" some random times... Even if packages are precompiled...

PR #716, opened 2018-08-19
Parameterization of options for `jupyter kernelspec install` in installkernel()

setempler:
Hi,

I had the problem to not be able to apply `IJulia.installspec` correctly for a configuration,
where a jupyter notebook server was installed in a python virtual environment (venv).

With python, it is possible to register a kernel installed in a venv into a server
that is also installed in a (different) venv. Key is to use the `--prefix` argument, such as in:

```bash
### bash

# install server in a python venv
export tmpd=$HOME/tmp/ijulia
mkdir -p $tmpd
python3 -m venv $tmpd # 3.7.0
source $tmpd/bin/activate
pip install jupyter
deactivate

# install a python kernel venv
python3 -m venv $tmpd/kernelenv
source $tmpd/kernelenv/bin/activate
pip install ipykernel
python -m ipykernel install \
    --name virtual_python \
    --prefix="$tmpd" \
    --display-name "Virtual Python"
deactivate

# verify
source $tmpd/bin/activate
jupyter kernelspec list
deactivate
```

So I request a pull for a change to allow customisation of the arguments
forwarded to `jupyter kernelspec install` to allow registering a Julia kernel
from `IJulia.installspec` to a server installed in a venv.

```julia
### julia

# install
envd = ENV["HOME"] * "/tmp/ijulia"
using IJulia
IJulia.installkernel("Virtual Julia", jupyter_options = ["--prefix='$envd'"])
```

```bash
### bash

# verify
source $tmpd/bin/activate
jupyter kernelspec list
deactivate
```

System:

* macOS 10.13.6
* bash 4.4.23
* python 3.7.0

PR #719, opened 2018-08-26
Parameterized version of installkernel

setempler:
This pull request relates to/is preferred over #716.

It extends the full control of `jupyter kernelspec install` parameters

* `name` (by keyword `"auto"` append version and debug tags, otherwise use name as is with spaces replaced to underscores)
* `display-name` (use as is instead of including a version and debug suffix)

It uses a different function name/definition to not break existing code.

Also it uses `vcat(...)` instead of `cat(..., dims)` to be backwards compatible w/ Julia 0.6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,25 @@ and a kernel called `Julia nodeps 0.6` (if you are using Julia 0.6)
 will be installed (will show up in your main Jupyter kernel menu) that
 lets you open notebooks with this flag.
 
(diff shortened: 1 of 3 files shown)


JamesWrigley:
Sorry for taking so long to get to this :see_no_evil: The display name can now be fully customized and the kernel name partially customized (it still adds the version number) so I'm going to close this.

PR #720, opened 2018-08-27
default to soft global scope

stevengj:
Defaults to "soft" global scope rules, as discussed in JuliaLang/julia#28789, to make it easier to write interactive code.  For example, the code
```jl
s = 0
for i = 1:10
    s += i
end
s
```
requires a `global` keyword in the `for` loop in Julia 1.0, but in IJulia this can be omitted.  Requires the [SoftGlobalScope package](https://github.com/stevengj/SoftGlobalScope.jl).

(The README gives opt-out instructions.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,14 @@ IJulia.set_max_stdio(1 << 20)
 The module that code in an input cell is evaluated in can be set using `Main.IJulia.set_current_module(::Module)`.
 It defaults to `Main`.
 
+### Opting out of soft scope
+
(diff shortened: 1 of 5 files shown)


michi-zuri:
Hi, sorry for this stupid question: I get how I can opt-out at run-time, but how can I do this when launching `jupyter notebook` from the terminal? How/Where should I include the environment variable to make all notebooks in that session use hard scope? 

> include the environment variable IJULIA_SOFTSCOPE=no environment of the IJulia kernel when it is launched.

stevengj:
You can include an environment variable in the kernel-configuration file.  We don't have an automated way to do this for you, so you need to edit the kernel configuration file yourself.

PR #721, opened 2018-08-29
Update get_token in handlers.jl to tokenize macros

rickhg12hs:
Change tokenizer logic so that char "@" can become part of the token.  Allows inspect_request to get info for macros.

### DIFF
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -234,11 +234,11 @@ function get_token(code, pos)
         end
     end
     endpos = startpos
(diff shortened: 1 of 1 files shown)


rickhg12hs:
Pretty sure this isn't the best tokenizer logic, but it extends the current `get_token` so that it can identify macros that start with `@`.  I don't know if this breaks anything else - I don't see any tests for this.

rickhg12hs:
@JamesWrigley 
> ... I can have a shot at it if that's ok with you. The branch will probably need rebasing since `get_token` was updated recently.

Please do.  Regards.

PR #731, opened 2018-09-04
Display the banner of Julia in the console ☺

aitzkora:
Not a great enhancement, but I wanted to see the julia banner when I used Julia in jupyter-console (because vim mode is not anymore supported since they remove readline).


aitzkora:
If I manage to find a way to hide the banner display in the console when you started the notebook, Il would be ok for you ?

aitzkora:
Eventually, I solved the problem putting ```Base.banner(IJulia.orig_stdout[])```  in my ```.julia/config/startup_ijulia.jl``` file. So we can close this issue

Issue #739, opened 2018-09-07
IJulia Installation Fails on Windows because ipykernel is not Installed


Kilo19:
I'm running Windows 10 1803. I tried to install IJulia with Julia 0.7.0 as well as 1.0.0 and both failed. Julia tries to intsall miniconda and tries to invoke the bundled Jupyter but could not find the Python executable. C:%homepath%\.julia\packages\IJulia\k5o7j\deps\build.log says:
```
Downloading and Extracting Packages
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... failed

ERROR conda.core.link:_execute(502): An error occurred while installing package 'defaults::ipykernel-4.9.0-py37_0'.
LinkError: post-link script failed for package defaults::ipykernel-4.9.0-py37_0
running your command again with `-v` will provide additional information
location of failed script: C:\Users\K19\.julia\packages\Conda\m7vem\deps\usr\Scripts\.ipykernel-post-link.bat
==> script messages <==
<None>
```
And then an ugly mess of exceptions:
```
[ Info: Installing Jupyter via the Conda package.
[ Info: Running `conda install -y jupyter` in root environment
[...]

PR #745, opened 2018-09-14
Check for startup_ijulia.jl and load it if present

### LINKED ISSUE #738: IJulia analog of atreplinit (detecting IJulia on startup)
If I put the following line in my `~/.julia/config/startup.jl` file:

```julia
const haveij = isdefined(Main, :IJulia)
```

then the value of `haveij` is `false`, even though `isdefined(Main, :IJulia)` obviously returns `true` when executed in a cell. What's the recommended way to customize initialization of your Julia session for IJulia?

timholy:
Fixes #738.

I should confess one oddity: for me, this works when I ask for a 0.7 kernel but it doesn't even seem to run when I ask for a 1.0 kernel. I've checked `pathof(IJulia)` from an IJulia cell with the 1.0 kernel, and it points to my `dev` directory. So I'm at a bit of a loss to explain the difference.


timholy:
I'm assuming the appveyor failures are unrelated?

stevengj:
Yes, the appveyor failures are #739.   

Needs some documentation in the README file.

I'm not sure what to make of what you say about 1.0 …

timholy:
Done.

I'm not sure either, but my hunch/hope is that it's an artifact of me developing it and that anyone who installs IJulia won't see the same weirdness. It really doesn't make any sense.

stevengj:
I tagged 1.12

stevengj:
It occurs to me that we should only load this if `Base.JLOptions().startupfile != 2`.  That is, don't load it if the user configured the kernel to pass `--startup-file=no`.

PR #748, opened 2018-09-24
Don't load startup_ijulia.jl with --startup-file=no

timholy:
ref https://github.com/JuliaLang/IJulia.jl/pull/745#issuecomment-423331925

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -9,7 +9,7 @@ end
 IJulia.init(ARGS)
 
 startupfile = abspath(homedir(), ".julia", "config", "startup_ijulia.jl")
-isfile(startupfile) && Base.include(Main, startupfile)
+isfile(startupfile) && Base.JLOptions().startupfile != 2 && Base.include(Main, startupfile)
 
 # import things that we want visible in IJulia but not in REPL's using IJulia
 import IJulia: ans, In, Out, clear_history


stevengj:
Thanks!

src/kernel.jl
```diff
@@ -9,7 +9,7 @@ end
 IJulia.init(ARGS)
 
 startupfile = abspath(homedir(), ".julia", "config", "startup_ijulia.jl")
-isfile(startupfile) && Base.include(Main, startupfile)
+isfile(startupfile) && Base.JLOptions().startupfile != 2 && Base.include(Main, startupfile)
```
stevengj: By the way, why `Base.include` and not just `include`?

timholy:
```julia
help?> include
search: include includet include_string include_dependency

  include(path::AbstractString)

  Evaluate the contents of the input source file in the global scope of the containing module. Every module (except those defined with baremodule) has its own 1-argument definition of include, which evaluates the file in that module. Returns the result of the last
  evaluated expression of the input file. During including, a task-local include path is set to the directory containing the file. Nested calls to include will search relative to that path. This function is typically used to load source interactively, or to combine
  files in packages that are broken into multiple source files.

  Use Base.include to evaluate a file into another module.
```
(see last line) and
```julia
julia> methods(include)
# 1 method for generic function "include":
[1] include(fname::AbstractString) in Base.MainInclude at client.jl:388

julia> methods(Base.include)
# 2 methods for generic function "include":
[1] include(mod::Module, path::AbstractString) in Base at sysimg.jl:23
[2] include(path::AbstractString) in Base at sysimg.jl:34
```

PR #755, opened 2018-10-07
Add custom MIME types to display_dict.

### LINKED ISSUE #749: Send arbitrary MIME types (without explicit call to display)
Similar to #595 but dropping the `display` with the explicit MIME type.

I.e. I'd like to be able to do something like

```js
showfoo(x::Foo) = "[ some representation of x ]"
# Hook up the text/vnd.foo MIME type?
display(x)
```

and have `display(x)` include the `text/vnd.foo` MIME type in the bundle.

For context, I'm trying to modify some widgets that requires a custom MIME type and don't want the user to have to do `display(MIME"application/vnd.mymime", widget)` all the time (since most of the time widgets are just implicitly displayed as a result of being the last expression evaluated in a cell).

Is that possible? It seems like it'd also avoid the `if ... elseif ... else if ... end if ... end` type chains that tend to appear everywhere around where MIME types are concerned. Possibly some kind of renderer-map (a dict from MIME to a renderer for that MIME type)?

twavv:
Fixes #749.

I think this is especially prudent given JupyterLab's focus on implementing custom MIME types rather than just rendering everything as HTML with JavaScript.

Very much open to comments (especially with regards to style and best practices!) and critiques.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -39,13 +39,15 @@ MIME types that when rendered (via stringmime) return JSON data. See
 This is necessary to embed the JSON as is in the displaydata bundle (rather than
 as stringify'd JSON).
 """
-const ijulia_jsonmime_types = Vector{Union{MIME, AbstractVector{MIME}}}([
+const ijulia_jsonmime_types = Vector{Union{MIME, Vector{MIME}}}([
     [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v3+json")],
     MIME("application/vnd.dataresource+json"),
 ])
 
-register_mime(x::Union{M, AbstractVector{M}}) where {M <: MIME} = push!(ijulia_mime_types, x)
-register_jsonmime(x::Union{M, AbstractVector{M}}) where {M <: MIME} = push!(ijulia_jsonmime_types, x)
+register_mime(x::Union{MIME, Vector{MIME}})= push!(ijulia_mime_types, x)
+register_mime(x::AbstractVector{<:MIME}) = push!(ijulia_mime_types, Vector{Mime}(x))
+register_jsonmime(x::Union{MIME, Vector{MIME}}) = push!(ijulia_jsonmime_types, x)
+register_jsonmime(x::AbstractVector{<:MIME}) = push!(ijulia_jsonmime_types, Vector{Mime}(x))
 
 include("magics.jl")
 
@@ -59,7 +61,7 @@ Generate the preferred MIME representation of x.
 Returns a tuple with the selected MIME type and the representation of the data
 using that MIME type.
 """
-function display_mimestring(mime_array::AbstractVector{MIME}, x)
+function display_mimestring(mime_array::Vector{MIME}, x)
     for m in mime_array
         if showable(mime_array, x)
             return display_mimestring(m, x)
@@ -76,7 +78,7 @@ Generate the preferred json-MIME representation of x.
 Returns a tuple with the selected MIME type and the representation of the data
 using that MIME type (as a `JSONText`).
 """
-function display_mimejson(mime_array::AbstractVector{MIME}, x)
+function display_mimejson(mime_array::Vector{MIME}, x)
     for m in mime_array
         if showable(mime_array, x)
             return display_mimejson(m, x)


travigd:
Ping @stevengj

Could you please take a look at this?

src/execute_request.jl
```diff
@@ -5,54 +5,111 @@
...
+IJulia will always try to include a "text/plain" representation of anything that
+is displayed. Since markdown and latex are specified within a sub-vector, IJulia
+will always try to render "text/markdown", and will only try to render
+"text/latex" if markdown isn't possible.
+"""
+const ijulia_mime_types = Vector{Union{MIME, Vector{MIME}}}([
+    MIME("text/plain"),
+    MIME("image/svg+xml"),
+    MIME("image/png"),
+    MIME("image/jpeg"),
```
stevengj: png and jpeg should be in a vector to have the same behavior as before

src/execute_request.jl
```diff
@@ -5,54 +5,111 @@
...
+
+"""
+MIME types that when rendered (via stringmime) return JSON data. See
+`ijulia_mime_types` for a description of how MIME types are selected.
+
+This is necessary to embed the JSON as is in the displaydata bundle (rather than
+as stringify'd JSON).
+"""
+const ijulia_jsonmime_types = Vector{Union{MIME, Vector{MIME}}}([
+    [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v3+json")],
```
stevengj: What about `application_vnd_dataresource`?

src/execute_request.jl
```diff
@@ -5,54 +5,111 @@
...
+])
+
+register_ijulia_mime(x::Union{MIME, Vector{MIME}}) = push!(ijulia_mime_types, x)
```
stevengj: No _ijulia_ here since that is already in the module name.  i.e. `IJulia.register_mime`

src/execute_request.jl
```diff
@@ -5,54 +5,111 @@
...
+will always try to render "text/markdown", and will only try to render
+"text/latex" if markdown isn't possible.
+"""
+const ijulia_mime_types = Vector{Union{MIME, Vector{MIME}}}([
+    MIME("text/plain"),
+    MIME("image/svg+xml"),
+    MIME("image/png"),
+    MIME("image/jpeg"),
+    MIME("text/html"),
+    [
```
stevengj: To be the same behavior as before, text/html should be in the list after text/markdown

travigd:
@stevengj I believe I've resolved all of those

src/execute_request.jl
```diff
@@ -5,54 +5,111 @@
...
-const image_png = MIME("image/png")
-const image_jpeg = MIME("image/jpeg")
-const text_markdown = MIME("text/markdown")
-const text_html = MIME("text/html")
-const text_latex = MIME("text/latex") # Jupyter expects this
-const text_latex2 = MIME("application/x-latex") # but this is more standard?
-const application_vnd_vega_v3 = MIME("application/vnd.vega.v3+json")
-const application_vnd_vegalite_v2 = MIME("application/vnd.vegalite.v2+json")
-const application_vnd_dataresource = MIME("application/vnd.dataresource+json")
+Base.showable(a::Vector{MIME}, x) = any(m -> showable(m, x), a)
```
stevengj: Should be `AbstractVector{<:MIME}`

travigd:
Done! :^)

src/execute_request.jl
```diff
@@ -20,7 +20,7 @@ is displayed. Since markdown and latex are specified within a sub-vector, IJulia
 will always try to render "text/markdown", and will only try to render
 "text/latex" if markdown isn't possible.
 """
-const ijulia_mime_types = Vector{Union{MIME, Vector{MIME}}}([
+const ijulia_mime_types = Vector{Union{MIME, AbstractVector{MIME}}}([
```
stevengj: No, this should be `Vector`

src/execute_request.jl
```diff
@@ -39,13 +39,13 @@ MIME types that when rendered (via stringmime) return JSON data. See
...
 """
-const ijulia_jsonmime_types = Vector{Union{MIME, Vector{MIME}}}([
+const ijulia_jsonmime_types = Vector{Union{MIME, AbstractVector{MIME}}}([
     [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v3+json")],
     MIME("application/vnd.dataresource+json"),
 ])
 
-register_mime(x::Union{MIME, Vector{MIME}}) = push!(ijulia_mime_types, x)
-register_jsonmime(x::Union{MIME, Vector{MIME}}) = push!(ijulia_jsonmime_types, x)
+register_mime(x::Union{M, AbstractVector{M}}) where {M <: MIME} = push!(ijulia_mime_types, x)
```
stevengj: This should probably be
```jl
register_mime(x::Union{MIME, Vector{MIME}}) = push!(ijulia_mime_types, x)
register_mime(x::AbstractVector{<:MIME}) = push!(ijulia_mime_types, Vector{MIME}(x))
```
in order to handle a wide range of argument types while keeping `ijulia_mime_types` a narrower type.

src/execute_request.jl
```diff
@@ -59,7 +59,7 @@ Generate the preferred MIME representation of x.
 Returns a tuple with the selected MIME type and the representation of the data
 using that MIME type.
 """
-function display_mimestring(mime_array::Vector{MIME}, x)
+function display_mimestring(mime_array::AbstractVector{MIME}, x)
```
stevengj: `AbstractVector{<:MIME}` would be broader as an argument type signature.   (However, since this is an internal function, `Vector{MIME}` is okay too.)

travigd:
:smile: Thanks for the guidance!

travigd:
Could you do a minor release so that I can depend on this in WebIO?

stevengj:
I was hoping to merge #756 too, but I will check back later today.

stevengj:
tagged 1.13

Issue #528, opened 2017-04-13
IOPub data rate exceeded, plotly


AshtonSBradley:
A new problem when using plotly backend in IJulia:
```julia
using Plots; plotly()
plot(real(sol[:,1]))
IOPub data rate exceeded.
The notebook server will temporarily stop sending output
to the client in order to avoid crashing it.
To change this limit, set the config variable
`--NotebookApp.iopub_data_rate_limit`.
```

So it looks like a default setting has been reduced. Unfortunately I can't see how to change this in Julia. Is there an easy way when launching IJulia from the REPL?

stevengj:
Apparently this is a bug in Jupyter 5, is that what you are using?  See ioam/holoviews#1181 for a similar problem and discussion of workarounds.

Apparently, this will be fixed in Jupyter 5.1: jupyter/notebook#2368

stevengj:
Until then, we could modify the `notebook()` function to launch with `jupyter notebook --NotebookApp.iopub_data_rate_limit=2147483647` if Jupyter 5 is detected.

bradleyashton:
using the version bundled with latest Julia 0.5.1 install (Conda):
```shell
grok:bin abradley$ ./jupyter --version
4.3.0
```

stevengj:
The data-rate limit was merged in February 2016: jupyter/notebook#931, so I guess it made it into Jupyter 4.3 as well, and maybe into Jupyter 4.2?  I'm surprised that no one has complained until now.

stevengj:
Okay, I've updated the `notebook()` command so that it launches Jupyter with the rate-limit disabled for 4.2 ≤ jupyter < 5.1.

juliohm:
Hi @stevengj, I am having the same issue with the GR backend now, can you please confirm the issue is not in IJulia anymore? https://github.com/JuliaPlots/Plots.jl/issues/1223

stevengj:
If you launch with `IJulia.notebook()` with Jupyter < 5.1, it passes `--NotebookApp.iopub_data_rate_limit=2147483647`.   If you run `jupyter` from a terminal manually, then you need to pass this yourself.

juliohm:
On the spot, thanks @stevengj

PR #791, opened 2019-01-03
don't require jupyter for installation

### LINKED ISSUE #692: Allow Installation Without Jupyter
I'm running Julia kernels in a dockerized environment (running kernels are connected to from frontends that are completely separate) where having Jupyter itself installed is completely superfluous (it adds a LOT of dependencies to install Jupyter + Conda when it won't even be used). Is there a way to just skip this step (e.g. environment variable)?

```
user@ef06f0d39e8b:/$ du -sh /usr/local/julia/share/julia/site/v0.6/Conda/
1.5G	/usr/local/julia/share/julia/site/v0.6/Conda/
```

stevengj:
Fixes #692.   Jupyter is only installed via Conda at the user's request when `IJulia.notebook()` runs.

Similarly, added an `IJulia.jupyterlab()` function; closes #783.  (Many thanks to @jordancluts, on whose code this is based.)

(`installkernel` also accepts a new `env` keyword to set the environment variables for a kernel.)

I removed the old Windows hacks for running Jupyter subcommands directly.  Hopefully they aren't needed anymore with Conda or any other functioning installation (i.e. you can just run `jupyter notebook` etcetera), since jupyter/notebook#448 was closed years ago?   I also removed the #528 workaround, since the corresponding Jupyter bug was fixed some time ago.

Since we nowadays require Julia 0.7, I dropped the Compat dependency.


stevengj:
Nope, I'm still getting `jupyter-notebook not found` on Windows.   I tried setting the PATH as in JuliaPy/Conda.jl#121 and it doesn't seem to help.  @tkf, any suggestions?

For now I guess we can go back to the hack of running the subcommand directly.  Running `jupyter-notebook.exe` works for me.

PR #793, opened 2019-01-06
Add Project.toml

jonathonf:
Julia [packages](https://docs.julialang.org/en/v1/stdlib/Pkg/) should have a Project.toml, 

> A package should have a project file with a uuid entry giving its package UUID. This UUID is used to identify the package in projects that depend on it.

This PR adds a `Project.toml` with the package's UUID and all [dependencies as of version 1.15](https://github.com/JuliaRegistries/General/blob/master/I/IJulia/Deps.toml).


stevengj:
Packages don't need to create the Project.toml file themselves… they are currently generated automatically, and eventually my understanding is that there will be automated pull requests to all registered packages once support for Julia 0.6 is dropped.

PR #796, opened 2019-01-18
Respect environment variables XDG_DATA_HOME and JUPYTER_DATA_DIR

oschulz:
IJulia.jl v1.15.0 broke Julia kernel installation on systems that don't/can't use the default Jupyter kernel directory.

This adds explicit support for `$XDG_DATA_HOME` and `$JUPYTER_DATA_DIR`.

See https://jupyter-core.readthedocs.io/en/latest/paths.html and
https://github.com/jupyter/jupyter_core/blob/master/jupyter_core/paths.py#L81 .

### DIFF
--- a/deps/kspec.jl
+++ b/deps/kspec.jl
@@ -15,15 +15,27 @@ copy_config(src, dest) = cp(src, joinpath(dest, basename(src)), force=true)
             (Ptr{Cvoid},Cint,Ptr{Cvoid},Cint,Ptr{UInt8}),C_NULL,CSIDL_APPDATA,C_NULL,0,path)
         return result == 0 ? transcode(String, resize!(path, findfirst(iszero, path)-1)) : homedir()
     end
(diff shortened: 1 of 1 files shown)


oschulz:
Bump. I'd be glad if a maintainer would be able to take a quick look at this - this issue is causing frequent trouble for us at the moment, it would be great if this (or a similar) fix could be included in the next release of IJulia.jl.

stevengj:
See also #800 — should we also be looking at `JUPYTER_PATH`?

I guess we should follow the logic here: https://github.com/jupyter/jupyter_core/blob/51185b86423a370e62455211fe051fc569472513/jupyter_core/paths.py

oschulz:
Thanks for the quick merge!

For future reference: As for `$JUPYTER_PATH`, I it has (to my knowledge) no influence on where Jupyter *installs* kernels, but only on where Jupyter will *search* for kernels.

PR #804, opened 2019-01-31
Implement message interrupt mode

### LINKED ISSUE #503: Interrupting the kernel does not work on Windows
I am running IJulia v1.4.1 on Julia v0.5.0 under Windows 10. 

I am unable to interrupt the kernel during a computation with Julia running on Windows, despite being able to interrupt the same operations without issue with Julia running on Linux. This is not a new issue - I have been using IJulia for about a year (mostly with Julia v0.4.x) and have never so far found the kernel interrupt feature to work on Windows.

In summary: The "interrupt" option in the "Kernel" menu appears to do nothing except flash the message "Interrupting the kernel" when the kernel is running on Windows.

filmor:
Fixes #503.

This switches entirely over to message mode. We can either make this configurable (though it must be set for Windows) or remove the signal mode completely.

/EDIT: After adjusting to the PR comments, message mode is only used on Windows.

### DIFF
--- a/deps/kspec.jl
+++ b/deps/kspec.jl
@@ -83,6 +83,8 @@ function installkernel(name::AbstractString, julia_options::AbstractString...;
             "display_name" => name * " " * Base.VERSION_STRING * debugdesc,
             "language" => "julia",
             "env" => env,
+            # Jupyter's signal interrupt mode is not supported on Windows
+            "interrupt_mode" => Sys.iswindows() ? "message" : "signal",
         )
 
         mkpath(juliakspec)
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -32,16 +32,17 @@ function eventloop(socket)
     end
 end
 
+const requests_task = Ref{Task}()
 function waitloop()
     @async eventloop(control[])
-    requests_task = @async eventloop(requests[])
+    requests_task[] = @async eventloop(requests[])
     while true
         try
             wait()
         catch e
             # send interrupts (user SIGINT) to the code-execution task
             if isa(e, InterruptException)
-                @async Base.throwto(requests_task, e)
+                @async Base.throwto(requests_task[], e)
             else
                 rethrow()
             end
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -261,6 +261,11 @@ function is_complete_request(socket, msg)
                            Dict("status"=>status, "indent"=>"")))
 end
 
+function interrupt_request(socket, msg)
+    @async Base.throwto(requests_task[], InterruptException())
+    send_ipython(requests[], msg_reply(msg, "interrupt_reply", Dict()))
+end
+
 const handlers = Dict{String,Function}(
     "execute_request" => execute_request,
     "complete_request" => complete_request,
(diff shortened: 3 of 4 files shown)


src/eventloop.jl
```diff
@@ -32,16 +32,17 @@ function eventloop(socket)
     end
 end
 
+requests_task = Ref{Task}()
```
stevengj: Should be `const`

src/handlers.jl
```diff
@@ -261,6 +261,11 @@ function is_complete_request(socket, msg)
                            Dict("status"=>status, "indent"=>"")))
 end
 
+function interrupt_request(socket, msg)
+    @async Base.throwto(requests_task[], InterruptException())
```
stevengj: Why is this `@async`?

deps/kspec.jl
```diff
@@ -83,6 +83,7 @@ function installkernel(name::AbstractString, julia_options::AbstractString...;
             "display_name" => name * " " * Base.VERSION_STRING * debugdesc,
             "language" => "julia",
             "env" => env,
+            "interrupt_mode" => "message",
```
stevengj: Seems like this should be `Sys.iswindows() ? "message" : "signal"`, since SIGINT is probably still more reliable on non-Windows systems.

stevengj:
I'm confused about how this is supposed to work.  If an `execute_request` is being handled, then in our current implementation the `eventloop` will be hung until that completes, so it will never get the `interrupt_request`, no?

src/handlers.jl
filmor: Because it blocks the control port otherwise indefinitely(?), at least that is what I observed. This is parallel to the construction in the main loop.

deps/kspec.jl
filmor: I don't really care much, though I don't see how there would be a difference in reliability. In the end, the same mechanism is used to interrupt the running `execute_request`.

filmor:
The message comes in via the control socket which is separate from the normal shell socket that gets the messages for execute, complete and other "normal" requests. Kernels are supposed to prioritise messages on the control socket.

stevengj:
Oh, I missed that — yes, we handle the control socket in a separate asynchronous task, so that should be fine.

In the current Julia release, tasks are scheduled cooperatively, so the execution task may be hogging all of the cycles if it refuses to yield (or do something like I/O that implicitly yields).  But in a future Julia release the tasks will run in separate threads (JuliaLang/julia#22631).

filmor:
I guess this is still the best we can do for Windows :). It works reliably (on both Windows and Linux) for a simple `sleep`, I haven't tried it yet with actual load.

deps/kspec.jl
stevengj: See below.  In the current Julia, the tasks are "green threads" that run cooperatively, so the task listening to the control socket may be starved of cycles.   sigint circumvents this.

stevengj:
(`sleep` lets other tasks run.)

deps/kspec.jl
```diff
@@ -83,6 +83,7 @@ function installkernel(name::AbstractString, julia_options::AbstractString...;
             "display_name" => name * " " * Base.VERSION_STRING * debugdesc,
             "language" => "julia",
             "env" => env,
+            "interrupt_mode" => Sys.iswindows() ? "message" : "signal",
```
stevengj: maybe put a `# see issue #503 — sigint doesn't work on Windows`

filmor:
Is there anything missing?

stevengj:
Thanks!

crstnbr:
Great to see this fixed! @stevengj, is this worth tagging a release?

stevengj:
Tagged 1.17.0

Issue #808, opened 2019-02-07
Kernel keeps dying with IJulia v1.15 or above


gdkrmr:
We are running a multiuser jupyterhub server as a service. With IJulia from v1.15 the kernel does not start. The error is: "UndefVarError: Compat not defined", traceback points to `/dir/to/IJulia/src/kernel.jl:2`. Version 1.14.1 works perfectly fine.

stevengj:
The current version of `kernel.jl` does not use or reference Compat at all.   Are you sure you updated it correctly?

gdkrmr:
Line 2 calls InteractiveUtils. And the entire file has not been changed for a while. So this may be a dependency issue.

stevengj:
InteractiveUtils is in the standard library and doesn’t use Compat. 

kernel.jl used to use Compat, so my first guess is that somehow you are using an old version of the file, maybe because of your file system setup?

stevengj:
The Compat usage was removed in September: https://github.com/JuliaLang/IJulia.jl/commit/34190dd16270c3a64a0e65cd508f0d5871b4a2e0

gdkrmr:
My bad, it was a custom kerne.json that used a kernel.jl file from an older version. Sorry for the noise.

gdkrmr:
maybe you could issue a warning in kernel.jl if the follwing check fails:
```julia
joinpath(dirname(pathof(IJulia)), "kernel.jl") == @__FILE__
```

stevengj:
Good idea.  A PR for that would be welcome.

gdkrmr:
No problem, I can do a PR. How would I make the warning appear inside the notebook, is a simple `@warn(...)` sufficient?

stevengj:
Wouldn't we want the warning to appear in the console, since it has to catch the case where the kernel.jl fails to load (in which case you don't have a running notebook)?    That is, just do `@warn ...`

gdkrmr:
Of course, you are right :-) the warning should appear in the console. Continuing the discussion in #809.

PR #809, opened 2019-02-09
test for the correct kernel.jl

gdkrmr:
Continuing the discussion from #808. 
- It works when you call `jupyter-notebook` from the terminal. 
- It does not work when you call `using IJulia; notebook()` inside a Julia process.

Does this need a unit test?


stevengj:
LGTM, thanks.

PR #813, opened 2019-02-21
Move mode and startup global variable into let block

davidanthoff:
Otherwise the name `mode` is used in the julia session that runs in a notebook. This is especially weird when one then does `using Distributions`, because message appears that `mode` is already defined in `Main`.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -43,8 +43,9 @@ Core.eval(Base, :(is_interactive = true))
 
 # check whether Revise is running and as needed configure it to run before every prompt
 if isdefined(Main, :Revise)
-    mode = get(ENV, "JULIA_REVISE", "auto")
-    mode == "auto" && IJulia.push_preexecute_hook(Main.Revise.revise)
+    let mode = get(ENV, "JULIA_REVISE", "auto")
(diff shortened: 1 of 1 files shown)


src/kernel.jl
```diff
@@ -43,8 +43,10 @@ Core.eval(Base, :(is_interactive = true))
 
 # check whether Revise is running and as needed configure it to run before every prompt
 if isdefined(Main, :Revise)
-    mode = get(ENV, "JULIA_REVISE", "auto")
-    mode == "auto" && IJulia.push_preexecute_hook(Main.Revise.revise)
+    let
+        mode = get(ENV, "JULIA_REVISE", "auto")
```
stevengj: Seems cleaner to write `let mode = ...` on the first line

davidanthoff: Done!

davidanthoff:
Btw, the same issue is probably true for `startupfile` [here](https://github.com/JuliaLang/IJulia.jl/blob/604c9a76ee8f079e856305e13f9c73cdf730771f/src/kernel.jl#L20). Should I put that into a `let` block as well? Or is that somehow different?

stevengj:
Yes, same for `startupfile`

davidanthoff:
Alright, `startup` is also in a `let` block now.

Issue #815, opened 2019-02-24
User_expressions are not evaluated, they're just returned untouched


bzinberg:
The Jupyter API's [`Execute`](https://jupyter-client.readthedocs.io/en/stable/messaging.html?highlight=user_expressions#execute) method has a field called `user_expressions` in both the `execute_request` and the `execute_reply`.

* The `execute_request.user_expressions` field is a dict with value type `display_data`, where each value represents an expression in the kernel's language that is to be evaluated.
* The `execute_reply.user_expressions` field is a dict with the same keys, and each value represents the result of evaluating the corresponding expression in the kernel.  The value type of this dict is not explicitly specified in the API doc, but should probably be either `display_data` or `execute_result` (which are very similar).

It appears that [this](https://github.com/JuliaLang/IJulia.jl/blob/fda9fe50950fd433d39f7bc7294e0de9376dd4cc/src/execute_request.jl#L82) line should be evaluating the expression `ex` in the kernel, but instead simply passes along `ex` itself.  (Also, I have encountered a separate issue where `parse` resolves to the wrong function, and I think the fix is to qualify it as `Meta.parse`; but I think this line needs to go away anyway).

The fix is to evaluate each of the input `user_expressions` in the kernel, and return the results in a dict of `execute_result`s.

PR #816, opened 2019-02-24
Fix Issue #815

bzinberg:
Fixes issue #815 by evaluating each of the `user_expressions` and returning them in an `execute_result`.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -79,8 +79,7 @@ function execute_request(socket, msg)
 
         user_expressions = Dict()
         for (v,ex) in msg.content["user_expressions"]
-            value = SOFTSCOPE[] ? softscope_include_string(current_module[], code) :
-                                  include_string(current_module[], code)
+            value = include_string(current_module[], ex)
             user_expressions[v] = Dict("data" => display_dict(value),
                                        "metadata" => metadata(value),
                                        "execution_count" => n)
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -92,7 +92,7 @@ function execute_request(socket, msg)
                 # The format of user_expressions[v] is like `error` except that
                 # it also has a `status` field:
                 # https://jupyter-client.readthedocs.io/en/stable/messaging.html#execution-errors
-                user_expressions[v] = Dict("status" => "ok",
+                user_expressions[v] = Dict("status" => "error",
                                            error_content(e, catch_backtrace())...)
             end
         end
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -79,14 +79,14 @@ function execute_request(socket, msg)
 
         user_expressions = Dict()
         for (v,ex) in msg.content["user_expressions"]
-	    try
-		value = include_string(current_module[], ex)
-		user_expressions[v] = Dict("data" => display_dict(value),
-					   "metadata" => metadata(value),
-					   "execution_count" => n)
-	    catch e
-		user_expressions[v] = error_content(e, catch_backtrace())
-	    end
+            try
+                value = include_string(current_module[], ex)
+                user_expressions[v] = Dict("data" => display_dict(value),
+                                           "metadata" => metadata(value),
+                                           "execution_count" => n)
+            catch e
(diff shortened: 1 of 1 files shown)


bzinberg:
Please advise on how best to test this.  I didn't see a unit test in `test/execute_request.jl` that looked like an obvious starting point for a test of the `user_expressions` functionality.  I did test it ad-hoc in the terminal, would a brief write-up of that be good enough?

src/execute_request.jl
```diff
@@ -79,7 +79,11 @@ function execute_request(socket, msg)
 
         user_expressions = Dict()
         for (v,ex) in msg.content["user_expressions"]
-            user_expressions[v] = invokelatest(parse, ex)
+            value = SOFTSCOPE[] ? softscope_include_string(current_module[], code) :
+                                  include_string(current_module[], code)
```
stevengj: I would just use `include_string` here.    Since the `user_expressions` here are a low-level thing used for communication between the front-end and the kernel, they shouldn't be affected by the users's `SOFTSCOPE` setting.

stevengj:
@minrk, I have to say that I find the documentation of `user_expressions` pretty vague.  Is the `execute_reply` supposed to be a dictionary of execute_reply-like contents, or display_data-like contents?  e.g. is it supposed to have an `execution_count` field?

src/execute_request.jl
stevengj: Shouldn't this be evaluating `ex`, not `code`?

stevengj:
@bzinberg, not that according to the docs we should check for exceptions, with something like:
```jl
try
    value = include_string(current_module[], ex)
    user_expressions[v] = Dict("data" => display_dict(value), "metadata" => metadata(value))
catch e
    user_expressions[v] = error_content(e, catch_backtrace())
end
```

src/execute_request.jl
bzinberg: Whoops, it absolutely should be evaluating `ex`.  And sounds good about blanked disabling softscope here.

bzinberg:
Thanks Steven for the review!  Incorporated your suggestions above.  Agreed that it would be great to clarify the API docs.  It might make sense to wait for that clarification before merging this PR.

bzinberg:
On second thought, I think we can reasonably decouple this PR from the task of updating the API docs (which should also happen, but without a blocking dependency).  We can do this by assuming the API allows whatever the IPython kernel does -- quoting [here](https://github.com/jupyter/jupyter/wiki/Jupyter-kernels):

> Kernel Zero is IPython, which you can get through ipykernel, and is still a
> dependency of jupyter. The IPython kernel can be thought of as a reference
> implementation, as CPython is for Python.

Here is what the IPython kernel does:

Via [this](https://github.com/ipython/ipykernel/blob/fe52dbd726cb405eb3a1e74e6a7fdb32372150ed/ipykernel/ipkernel.py#L331) line, the content of the `user_expressions` is the output of [`user_expressions(...)`](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2617-L2642).

For `user_expressions` that evaluate successfully, the formatting [is](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2638) controlled by `_format_user_obj`, which [produces](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2609-L2614) something that looks like a [`display_data`](https://jupyter-client.readthedocs.io/en/stable/messaging.html?highlight=user_expressions#display-data) except that it also has a `status` field.

For `user_expressions` that produce an exception, the formatting [is](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2640) controlled by [this](https://github.com/ipython/ipython/blob/master/IPython/core/interactiveshell.py#L2585-L2601) function which returns an exception object with fields `status, ename, evalue, traceback` as described in the [API doc](https://jupyter-client.readthedocs.io/en/stable/messaging.html?highlight=user_expressions#execute).

I have updated the IJulia implementation to produce these formats.  @stevengj, how does this sound to you?

stevengj:
LGTM.

PR #821, opened 2019-03-08
fix(vega): add v4 mime type

mcmcgrath13:
The v4 mime type is what is currently supported out of the box by Jupyter, v3 requires an extension to be installed.

This PR adds the v4 mime type to the lists of supported mime types.  MIME"application/vnd.vega.v3+json" was left for backwards compatibility.  Vega recently released v5, so I expect this will need to be updated again in the near future.

I am concurrently putting in a PR for [VegaLite.jl](https://github.com/fredo-dedup/VegaLite.jl) to update the mime type on that end too.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -35,7 +35,7 @@ This is necessary to embed the JSON as is in the displaydata bundle (rather than
 as stringify'd JSON).
 """
 const ijulia_jsonmime_types = Vector{Union{MIME, Vector{MIME}}}([
-    [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v3+json")],
+    [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v4+json"), MIME("application/vnd.vega.v3+json")],
     MIME("application/vnd.dataresource+json"), MIME("application/vnd.plotly.v1+json")
 ])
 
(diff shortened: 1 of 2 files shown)


stevengj:
This is getting little ridiculous.  Why don't they put the version number into the JSON content rather than continually updating the MIME type?

src/display.jl
```diff
@@ -35,7 +35,7 @@ This is necessary to embed the JSON as is in the displaydata bundle (rather than
 as stringify'd JSON).
 """
 const ijulia_jsonmime_types = Vector{Union{MIME, Vector{MIME}}}([
-    [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v3+json")],
+    [MIME("application/vnd.vegalite.v2+json"), MIME("application/vnd.vega.v4+json"), MIME("application/vnd.vega.v3+json")],
```
stevengj: Would be cleaner to do:
```jl
[MIME("application/vnd.vegalite.v$n+json") for n=2:4]
```
and similarly for `ipy_mime`

mcmcgrath13:
@stevengj Ironically, they do both... It looks like VegaLite also just released v3. So it probably makes sense to do the comprehension for both mime types vegalite from 2:3 and vega from 3:5.  I'll get that added.

PR #834, opened 2019-04-15
Fix `IJulia not found` when installed, but not in user's Environment

NHDaly:
Add the IJulia package that contains kernel.jl as last element of `LOAD_PATH`.  If IJulia isn't found in the user's environment, this will then default to loading the version of IJulia that the kernel.jl was installed from.

This most commonly occurs when the user has installed IJulia in a custom environment, but not the top-level shared environment.
This is a fix for the issue discussed here: https://github.com/JuliaLang/IJulia.jl/pull/820#issuecomment-482839448

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -1,3 +1,7 @@
+# Manually add this package as the last entry in the LOAD_PATH so that if IJulia isn't
+# installed in the user's Environment, IJulia will still be picked up from this directory.
+push!(LOAD_PATH, dirname(@__DIR__))
+
 import IJulia
 
 let


NHDaly:
Bumping this PR, which I had let languish. @stevengj, would you be able to have a look.

Coming back to it again after a few years, I think this seems like a sane PR to me. I know that i've experienced this issue before when switching between julia environments, and i think this would be a nice simplification for users who install IJulia from a local environment and then attempt to open a Julia kernel directly via the Jupyter UI later.

stevengj:
Manually adding IJulia to the load path seems chancy to me.   What if ZMQ is not in the environment either?  Or an incompatible version of ZMQ?

NHDaly:
🤔 yeah, makes sense! thanks for the thoughts. I'm happy to close it then. I think other approaches like automatically installing into the top-level Environment could be preferable. Thanks 👍 closing

PR #846, opened 2019-06-02
Use DEPOT_PATH[1]... to load startup_ijulia.jl

macd:
Use DEPOT_PATH[1] instead of homedir()/.julia to load startup_ijulia.jl

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -22,7 +22,7 @@ ENV["COLUMNS"] = get(ENV, "COLUMNS", 80)
 
 IJulia.init(ARGS)
 
-let startupfile = abspath(homedir(), ".julia", "config", "startup_ijulia.jl")
+let startupfile = abspath(DEPOT_PATH[1], "config", "startup_ijulia.jl")
     isfile(startupfile) && Base.JLOptions().startupfile != 2 && Base.include(Main, startupfile)
 end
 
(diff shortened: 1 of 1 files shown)


src/kernel.jl
```diff
@@ -22,7 +22,7 @@ ENV["COLUMNS"] = get(ENV, "COLUMNS", 80)
 
 IJulia.init(ARGS)
 
-let startupfile = abspath(homedir(), ".julia", "config", "startup_ijulia.jl")
+let startupfile = abspath(DEPOT_PATH[1], "config", "startup_ijulia.jl")
```
stevengj: The whole thing should be wrapped in `!isempty(DEPOT_PATH)` similar to how [`startup.jl` is loaded](https://github.com/JuliaLang/julia/blob/5d02c59185ef31703891adc748e0ed414a9a5b3f/base/client.jl#L327)

src/kernel.jl
```diff
@@ -22,7 +22,7 @@ ENV["COLUMNS"] = get(ENV, "COLUMNS", 80)
 
 IJulia.init(ARGS)
 
-let startupfile = abspath(homedir(), ".julia", "config", "startup_ijulia.jl")
+let startupfile = !isempty(DEPOT_PATH) ? abspath(DEPOT_PATH[1], "config", "startup_ijulia.jl") : ""
```
stevengj: If the depot path is empty, I don’t think we should try to load a startup file at all

macd:
Agreed, but if the DEPOT_PATH is empty, then startupfile = "" and
isfile(startupfile) returns false so no attempt is made to load a startup
file.  I guess the logic is not as clear as it could be, but was just doing
minimum edit from what was already there.

PR #847, opened 2019-06-12
Improved tooltip: functions, tuples, etc

### LINKED ISSUE #631: Improve tooltips by always showing the docstring of the function being called
## IPython 

Pressing `shift + tab` when calling a function `f` displays the docstring of `f`:

python2

## IJulia

The behaviour is different in IJulia. If a number is given as the first argument, then no tooltip is displayed:

julia2

However, if a defined variable `x` is supplied, then the docstring of `x` is shown:

julia3

Shouldn't we make the tooltip always show the docstring of the function being called?

rongcuid:
This is an attempt to fix #631. 

I have implemented the behavior specified in the issue, allowing tooltip for `f` when Shift-Tab is pressed inside a function call `f(x, y, ... |)` or right outside the call `f(x, y, ...)|`. `|` is cursor. Detailed behavior is in docstring.

The current implementation has some string hacks which I consider ugly (explained thoroughly, I hope, in docstrings). If I can dig deeper, I prefer an implementation using a proper Julia language parser. Maybe one that comes from Julia itself. 

In any case, I tested with quite a few of my Notebooks, and they all seem to work well. 

I forgot to squash the commits, though.

### DIFF
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -246,22 +246,39 @@ end
 docdict(s::AbstractString) = display_dict(Core.eval(Main, helpmode(devnull, s)))
 
 import Base: is_id_char, is_id_start_char
-function get_token(code, pos)
-    # given a string and a cursor position, find substring to request
-    # help on by:
-    #   1) searching backwards, skipping invalid identifier chars
-    #        ... search forward for end of identifier
-    #   2) search backwards to find the biggest identifier (including .)
-    #   3) if nothing found, do return empty string
-    # TODO: detect operators?
 
-    # When autocompletion is enabled and the user is going down the list of the
-    # completions, `pos` may become out of bounds. In this case we set it back
-    # to the last valid index.
-    pos = min(pos, lastindex(code))
+"""
+    get_previous_token(code, pos, crossed_parentheses)
 
+Given a string and a cursor position, find substring corresponding to previous token.
+`crossed_parentheses:Int` keeps track of how many parentheses have been crossed.
+A pair of parentheses yields 0 crossing; a '(' add 1; a ')' subtracts 1.
+
+Returns `(startpos, endpos, crossed_parentheses, stop)`
+
+- `startpos` is the start position of the closest potential token before `pos`.
+- `endpos` is end position if said token is can be valid identifier, or `-1` otherwise
+- `crossed_parentheses` is the new count for parentheses.
+- `stop` is true if ';' is hit, denoting the beginning of a clause.
+"""
+function get_previous_token(code, pos, crossed_parentheses)
     startpos = pos
+    separator = false
+    stop = false
     while startpos > firstindex(code)
+        c = code[startpos]
+        if c == '('
+            crossed_parentheses += 1
+            selarator = false
+        elseif c == ')'
+            crossed_parentheses -= 1
+            separator = false
+        elseif c == ';'
+            stop = true
+        elseif !is_id_char(c) && !isspace(c) && !separator
+            separator = true
+            crossed_parentheses = max(0, crossed_parentheses - 1)
+        end
         if is_id_char(code[startpos])
             break
         else
@@ -274,7 +291,7 @@ function get_token(code, pos)
     end
     startpos = startpos < pos ? nextind(code, startpos) : pos
     if !is_id_start_char(code[startpos])
-        return ""
+        return startpos, -1, crossed_parentheses, stop
     end
     while endpos < lastindex(code) && is_id_char(code[endpos])
         endpos = nextind(code, endpos)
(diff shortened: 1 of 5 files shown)


src/handlers.jl
```diff
@@ -190,17 +190,39 @@ end
...
+    - `stop` is true if ';' is hit, denoting the beginning of a clause.
+    """
     startpos = pos
+    separator = false
+    stop = false
     while startpos > firstindex(code)
+        c = code[startpos]
+        if c == '('
+            crossed_parentheses += 1
+            selarator = false
```
stevengj: separator?

stevengj:
Would be good to add a `test/completion.jl` with some tests for `get_token`.

src/handlers.jl
```diff
@@ -190,17 +190,39 @@ end
...
+        elseif c == ')'
+            crossed_parentheses -= 1
+            separator = false
+        elseif c == ';'
+            stop = true
+        elseif !is_id_char(c) && !isspace(c) && !separator
+            separator = true
```
stevengj: Shouldn't this just be `c == ','` or `c in (',', ...some set....)`?   There are lots of things that aren't ID chars that might be encountered in Julia code, e.g. `@` and `{` or any binary operator.

src/handlers.jl
```diff
@@ -213,15 +235,84 @@ function get_token(code, pos)
...
+
+    - `f()|` has balanced parentheses with nothing within, thus `f` is the desired token.
+    - `f(|)` has imbalanced parentheses, thus `f` is the desired token.
+    - `f(x|, y)` gives tokens `x` and `f`. `x` has balanced parentheses, while `f` is left-imbalanced. `f` is desired.
+    - `f(x)|` returns `f`
+    - `f(x, y)|` returns `f`.
+    - `f((x|))` returns `f`, as expected
+    - `f(x, (|y))` returns `f`. **This is a hack**, as I deduct `crossed_parentheses` whenever a separator is encountered, clamped to 0!
+        Otherwise, `x` would be returned.
+    - `f(x, (y|))`, `f(x, (y)|)`, and `f(x, (y))|` all behave as above. Arbitrary nesting of tuples should not cause misbehavior.
```
stevengj: What do we do if the argument is a more complicated expression, e.g. `f("foo bar")` or `f(3 + 5 -2)`?   Or `Vector{Int}(undef, n)`?

stevengj:
I agree that it would be nice to use the Julia parser; the basic issue is that the user might be asking for help on code they haven't finished writing yet…

src/handlers.jl
rongcuid: The "separator" is the hack I referred to in my comments. It basically handles special cases like this:

```
f(..., X, (|Y, ...))
```

On encountering anything not a parenthesis and not an identifier, my algorithm deducts parentheses count. Without the hack, the token returned would be `X` instead of `f`. It is an incredibly bad hack which is definitely not the correct algorithm, but it works in a wide variety of cases and is at least better than the original implementation which can only return the token immediately before cursor.

src/handlers.jl
rongcuid: It might actually better to use your conditional cases. I was cutting shortcuts...

src/handlers.jl
rongcuid: `f("foo bar")` always returns `f` no matter where the cursor is (tested interactively).

`f(3 + 5 - 2)` always returns `f` no matter where the cursor is (also tested interactively).

`Vector{Int}(undef, n)` _happens to_ return `Vector` no matter where the cursor is. I did not account for `{}` in code, but it just happens to work.

rongcuid:
I will try quickly adding some test cases. It's the first time I develop Julia packages so I will need to check out some docs...

rongcuid:
I have changed behavior slightly to make sure the basic `x + y + z` works, giving `x`, `y`, and `z` respectively. Further making the completion more robust might be out of scope for my goals, so I propose to merge this as an intermediate solution which provides a slight quality of life improvement. It should be good enough for many common cases.

rongcuid:
As a note, the julia tokenizer might be helpful for a better solution:

[](https://github.com/JuliaLang/Tokenize.jl)

rongcuid:
Do you have any plan on merging? I think the current state is good enough for an intermediate improvement.

rongcuid:
Any updates?

rongcuid:
Any activities on this? @stevengj can this be merged, or are there some feedback?

src/handlers.jl
```diff
@@ -190,17 +190,39 @@ end
...
-    # given a string and a cursor position, find substring to request
-    # help on by:
-    #   1) searching backwards, skipping invalid identifier chars
-    #        ... search forward for end of identifier
-    #   2) search backwards to find the biggest identifier (including .)
-    #   3) if nothing found, do return empty string
-    # TODO: detect operators?
 
+function get_previous_token(code, pos, crossed_parentheses)
+    """
```
fredrikekre: If you want this to be a docstring it should be placed before the function definition, otherwise, keep it as a comment. Right now it is just an unecessary string in the function body.

rongcuid: Fixed

rongcuid:
Any more reviews?

rongcuid:
Do you think this is mergable or is it too hacky to be put into mainline? I have been waiting for a decision for quite a while.

rongcuid:
Is there anything I still need to do? Or is this superseded by some other patches?

JamesWrigley:
Sorry for letting this languish @rongcuid, would you still be able to rebase this? If so then I'll review it. And if not, would it be ok if I finished off the PR?

rongcuid:
I won't be able to rebase this as I am not actively using IJulia now. You can close it

JamesWrigley:
Thanks @rongcuid, I took the liberty of rebasing and updating the code a bit. This seems like a solid improvement over what we have right now so I'm going to go ahead and merge it. We might revisit it later to see if Tokenize.jl can be used instead.

src/handlers.jl
```diff
@@ -246,22 +246,39 @@ end
...
+"""
+function get_previous_token(code, pos, crossed_parentheses)
     startpos = pos
+    separator = false
+    stop = false
     while startpos > firstindex(code)
+        c = code[startpos]
+        if c == '('
+            crossed_parentheses += 1
+            selarator = false
```
halleysfifthinc, line 272: This typo was never fixed

```suggestion
            separator = false
```

stevengj: https://github.com/JuliaLang/IJulia.jl/pull/1171

JamesWrigley: Oops, thanks for catching that :grimacing:

PR #859, opened 2019-07-29
Add `display_payload` function.

twavv:
This allows users (mostly package authors) to specify the data and metadata in the same method.

This is useful for my work on JuliaPy/PyInteract.jl#1 because we want to hook into the way that IPython generates the display information, but IPython works by returning a tuple (so to generate both we'd have to call the IPython formatter twice which is less than ideal).

Constructive criticism gladly encouraged. :^)

Ping @stevengj.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -87,6 +87,9 @@ display_mimejson(m::MIME, x) = (m, JSON.JSONText(limitstringmime(m, x)))
 Generate a dictionary of `mime_type => data` pairs for all registered MIME
 types. This is the format that Jupyter expects in display_data and
 execute_result messages.
+
+This function may also return `nothing`, in which case no Jupyter message is
+sent at all.
 """
 function display_dict(x)
     data = Dict{String, Union{String, JSONText}}()
@@ -114,8 +117,23 @@ function display_dict(x)
     end
 
     return data
-
 end
+
+display_dict(::Nothing) = nothing
+
+"""
+Generate a `(data=..., metadata=...)` tuple.
+
+This defaults to using `display_dict` and `metadata` respectively (see the
+documentaiton for those functions to see what `data` and `metadata` should be).
+A custom method for this function can be specified whenever it is difficult to
+compute the data and metadata separately. Most users will just need to define a
+method for `display_dict`.
+
+This function may also return `nothing`, in which case no Jupyter message is
+sent at all.
+"""
+display_payload(x) = (data=display_dict(x), metadata=metadata(x))
 
 # queue of objects to display at end of cell execution
 const displayqueue = Any[]
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -107,16 +107,23 @@ function execute_request(socket, msg)
         undisplay(result) # dequeue if needed, since we display result in pyout
         invokelatest(display) # flush pending display requests
 
-        if result !== nothing
-            result_metadata = invokelatest(metadata, result)
-            result_data = invokelatest(display_dict, result)
-            send_ipython(publish[],
-                         msg_pub(msg, "execute_result",
-                                 Dict("execution_count" => n,
-                                              "metadata" => result_metadata,
-                                              "data" => result_data)))
-
+        result_payload = invokelatest(display_payload, result)
+        if result_payload !== nothing && result_payload[1] !== nothing
+            data, metadata = result_payload
+            send_ipython(
+                publish[],
+                msg_pub(
+                    msg,
+                    "execute_result",
+                    Dict(
+                        "execution_count" => n,
+                        "metadata" => metadata,
+                        "data" => data,
+                    ),
+                ),
+            )
         end
+
         send_ipython(requests[],
                      msg_reply(msg, "execute_reply",
                                Dict("status" => "ok",


travigd:
Note: I also special-cased returning nothing so that things other than literal `nothing` can also avoid outputting an `Out[123]` in the notebook (this is what ipywidgets does). That part isn't super important to me though (we could just mandate that you always return a Dict).

src/display.jl
```diff
@@ -114,9 +117,24 @@ function display_dict(x)
...
+This defaults to using `display_dict` and `metadata` respectively (see the
+documentaiton for those functions to see what `data` and `metadata` should be).
+A custom method for this function can be specified whenever it is difficult to
+compute the data and metadata separately. Most users will just need to define a
+method for `display_dict`.
+
+This function may also return `nothing`, in which case no Jupyter message is
+sent at all.
+"""
+display_payload(x) = (data=display_dict(x), metadata=metadata(x))
```
stevengj: This is fine with me in principle, but I'd prefer to separate it from IJulia somehow, so that modules that want to support this don't need to pull in an IJulia dependency.

Maybe we can define it as a `repr` method with a custom MIME type?  e.g. you define
```jl
Base.repr(::MIME"application/x-ijulia-payload", x::MyObject) = (data=display_dict(x), metadata=metadata(x))
```

travigd: Yeah absolutely I see what you mean. I'm not hugely a fan of hacking `repr` with a custom MIME (everything else to my knowledge returns a string which I think is more in-line with how MIMEs are meant to work), but without a suitable alternative, I think it's... fine. Alternatively we could just tell them to use Requires.jl (which is what things like WebIO use already).

Also just another good use case for better integrated optional dependencies :^). It seems like Julia's package ecosystem isn't great right now for package relations that arent a very strong "A depends on B".

stevengj: Better to create a small package that both IJulia and other packages could depend on.

travigd: MIMEBundle.jl?

stevengj: Sure.

stevengj: Probably `MIMEBundles.jl` for the package and module, and `MIMEBundle` for a type declared in the module.

travigd: Where should this package live? I can volunteer the Gizmos org (a lot of that org is already concerned with rich content so it makes sense) but wherever is fine with me.

stevengj: Probably the same org as IJulia, i.e. JuliaLang.   Or maybe IJulia should be moved to JuliaGizmos?

stevengj: I guess we can put it in JuliaGizmos

travigd: I don't feel super strongly, I just think that we should put it in a place that makes sense. If you'd like me to create the package in JuliaGizmos, I can do that and add some starter code.

stevengj: Sure, please gobahead

travigd: https://github.com/JuliaGizmos/MIMEBundles.jl

Wrote some initial stuff. I also took an opportunity to clean-up some API (but we should be able to forward the current IJulia functions to the equivalent for MIMEBundles).

travigd:
@stevengj Are you still interested in all of this stuff? There are a few (minor) changes to be made to IJulia, including using MIMEBundles.jl, to make ipywidgets + PyCall work seamlessly.

I think I'd be able to do the work it would just require some buy-in and a bit of review from you. :^)

stevengj:
It seems like the MIMEBundles package should be more minimal.   Shouldn't it be just for people who want to have a specific precomputed set of MIME representations for their object?

* It seems like the `mimerepr` stuff should stay in IJulia, since how that works is kind of specific to IJulia. 

* All of the `DEFAULT_MIMES` stuff should also probably stay in IJulia.

On the other hand

* There should probably also be a `show(mime, mimeobject)` method to output the MIMEBundle object as a given type based on the stored data.   So things like the REPL would continue to display the wrapped object as if you had displayed it directly?

travigd:
To be clear, I'm not really advocating for this PR in particular, it's just that this PR is where all the discussion of MIMEBundles got started.

> It seems like the MIMEBundles package should be more minimal. Shouldn't it be just for people who want to have a specific precomputed set of MIME representations for their object?

It currently is pretty minimal: https://github.com/JuliaGizmos/MIMEBundles.jl.

The current use (by a third party package) would be
```julia
struct MyType ... end
data = MyType()

# Use default MIMEBundle, which includes common MIMEs.
default_bundle = MIMEBundle(data)

# Define our own overload
function MIMEBundle(data::MyType)
    return MIMEBundle(
        Dict("text/plain" => "MyType(...)", ...),
        Dict(), # empty metadata
    )
end
custom_bundle = MIMEBundle(data)
```

Looking at this, I think I'd want to move away from defining overloads for `MIMEBundle` and instead have a `mimebundle` function which will generate a `MIMEBundle` object (to avoid ambiguous methods with the `MIMEBundle` constructor). But that's a detail™.

> It seems like the mimerepr stuff should stay in IJulia, since how that works is kind of specific to IJulia.

> All of the DEFAULT_MIMES stuff should also probably stay in IJulia.

I'm not sure of a good way to do this, honestly. Currently, I have [this](https://github.com/JuliaGizmos/MIMEBundles.jl/blob/master/src/mimes.jl).

One possible option would be to add a trait-esque way of defining custom methods on `MIMEBundle`.

```julia
data = ...
# Default constructor
bundle = MIMEBundle(data)

# Specific constructor for IJulia
# Custom types would overload MIMEBundle(::IJuliaMIMEs, data::MyType)
ijulia_bundle = MIMEBundle(IJuliaMIMEBundle(), data)

# If no overload, MIMEBundles.jl would do something like this behind the scenes
function MIMEBundle(bundle_strategy::AbstractBundleStrategy, data)
    # IJulia would define default_mimes(::IJuliaMIMEs) to be what's currently defined as the
    # default MIME list.
    mimes = default_mimes(bundle_strategy)
    # Return a bundle that generates that has all of the requested MIMEs (if applicable for the data type)
end
```

If I define `MIMEBundle(::MyType)` (even without defining `MIMEBundle(::IJuliaMIMEs, ::MyType)`), it would probably use the defined method instead of using the default IJulia MIMEs.

> There should probably also be a show(mime, mimeobject) method to output the MIMEBundle object as a given type based on the stored data. So things like the REPL would continue to display the wrapped object as if you had displayed it directly?

I can add the overloads for `Base.show` in MIMEBundles.jl.

travigd:
@stevengj Any thoughts on any of this? We started [PyInteract.jl](https://github.com/JuliaPy/PyInteract.jl) many moons ago, and this is the last piece of that puzzle.

stevengj:
Am traveling now; please ping me again next week

travigd:
@stevengj It has been >= 1 week.

djsegal:
I've been following this chain of events for some time:
+ https://github.com/JuliaPy/PyInteract.jl/issues/1
+ https://github.com/JuliaPy/PyCall.jl/issues/89

And am still very interested in the prospect of using widgets at the ground level. Let me know if you guys need any help with this

Issue #862, opened 2019-08-05
IJulia doesn't work with Jupyter --transport=ipc 


marius311:
I'm trying to install a Julia 1.1 / IJulia 1.19.0 kernel on JupyterLab 1.0.4. It installs fine, but then the kernel hangs whenever I try to evaluate anything. The JupyterLab statusbar tells me `Julia 1.1 | Connected`. I have tried enabling IJulia debug but nothing is printed. By inserting my own print statements, I have verified that it at least makes it to right before the line: https://github.com/JuliaLang/IJulia.jl/blob/a93ddfdd834a3cbf7fb1990f7f77109e700a8563/src/kernel.jl#L52

One weird thing is that this is happening when JupyterLab is launched by JupyterHub, but on the same system if I launch JupyterLab by hand, the kernel seems to work fine. 

Any suggestions how I can figure out what is going on? Thanks for any help.

rcthomas:
I've confirmed this behavior on the same system.  I also have a docker-compose setup that runs a hub and a notebook container with 2 test users in it.  If you install IJulia into a test user account on the target container and try to use the kernel, you get the behavior described by @marius311 here.

I've set logging on the notebook config to report everything, and turned on `c.Session.debug`.  So far mostly I see messages from the notebook server like 

`[W 2019-08-05 22:11:40.485 SingleUserLabApp handlers:229] Timeout waiting for kernel_info reply from e0a618ab-d82c-48f8-a8f9-1bb297a733f6`

In another session I saw (info lines in between redacted):

```
[W 2019-08-05 22:07:03.572 SingleUserLabApp handlers:229] Timeout waiting for kernel_info reply from c05c1303-1c69-435a-a3e8-b0bf51e2d51d
[E 2019-08-05 22:07:03.576 SingleUserLabApp handlers:276] Error opening stream: HTTP 404: Not Found (Kernel does not exist: c05c1303-1c69-435a-a3e8-b0bf51e2d51d)
[E 2019-08-05 22:07:03.588 SingleUserLabApp handlers:276] Error opening stream: HTTP 404: Not Found (Kernel does not exist: c05c1303-1c69-435a-a3e8-b0bf51e2d51d)
```

Finally, when I shut down the server from the hub, I get a bunch of json output from the notebook server log that starts with

```
{'content': {},
 'header': {'date': datetime.datetime(2019, 8, 6, 5, 10, 40, 548133, tzinfo=datetime.timezone.utc),
[...]

PR #868, opened 2019-08-31
make work with ipc transport

marius311:
Fixes https://github.com/JuliaLang/IJulia.jl/issues/862

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -84,11 +84,12 @@ function init(args)
     requests[] = Socket(ROUTER)
     control[] = Socket(ROUTER)
     heartbeat[] = Socket(ROUTER)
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks!  (See also https://github.com/ipython/ipykernel/commit/6e60f7259352ffd3fed184a22d49e48ded1787af)

@minrk, I noticed that [ipykernel has a more complicated logic](https://github.com/ipython/ipykernel/blob/ca3077e87618eacbb5e312af3f10935269ed91e2/ipykernel/kernelapp.py#L183-L200) for detecting port conflicts.  I guess we should be doing something similar?  I've never run into this problem myself.

aramirezreyes:
Can we tag a new version?

PR #876, opened 2019-10-05
Add InlineIOContext with Jupyter property.

travigd:
Adds a special property to the `IOContext` used when displaying data (as well as a helper function to create aforementioned `IOContext`).

My use case for this is wanting to avoid double-serializing things when using Jupyter and WebIO. I want to be able to detect that we're using IJulia to display and therefore not render the full HTML structure (which is actually rendered using a custom JSON MIME type).

Included test case for usage.


travigd:
Ping @stevengj

travigd:
Note: this is adjacent to #859 but both can exist in harmony. :^)

travigd:
Bump

Issue #902, opened 2020-02-11
How to fix blurry PyPlot plots in HiDPI machine in jupyter?


mistguy:
In a hidpi machine, the inline plots of jupyter via PyPlot.jl are blurry.

Previously, I usually use the PyPlot.svg(true) option. However, this will slow down the browser considerably if the plot is really large.

I noticed in a jupyter python session, one can do %config InlineBackend.figure_format = 'retina' to fix this issue. However, since %config is not supported by IJulia, how can I set the figure_format option?

This question was posted on https://discourse.julialang.org/t/how-to-fix-blurry-pyplot-plots-in-hidpi-machine-in-jupyter/33471 but was not getting any attention.

stevengj:
Basically, someone needs to dig through the Jupyter code to see what this `%config` option actually does…

stevengj:
Relevant code:
* https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/pylabtools.py#L138-L147
* https://github.com/ipython/ipython/blob/d262f0c4bda68e07de3c20be78606f13fc4bcba1/IPython/core/display.py#L1074-L1305
* https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/pylabtools.py#L112-L114

 Basically, if this setting is true, it doubles the DPI, and then returns some corresponding metadata to the front-end.

xzackli:
As a temporary hack, I've found success on OSX by just applying a little CSS with Stylebot, a Chrome browser extension that lets you apply arbitrary CSS rules to pages. The css just scales the image by 2x.

```css
.output_subarea.output_png > img {
    zoom: 50%;
}
```

PR #918, opened 2020-04-21
handle retina/hidpi displays with metadata

xzackli:
This implements the handling of hidpi displays for PyPlot (#902) like IPython does. During `display`, it sends metadata with the PNG images that specify half-height and width, with these dimensions read from the PNG header. See the issue for links to how IPython does it.

To complete this feature, it might be nice to add a `PyPlot.retina(true)` command which doubles the inline DPI and sets `IJulia.retina[]`.

### Example Usage
```julia
using IJulia
IJulia.retina[] = true

using PyPlot
plt.plot( sin.(1:0.1:10) )
```


src/inline.jl
```diff
@@ -88,14 +88,36 @@ function display(d::InlineDisplay, M::MIME, x)
...
 end
 
+
+
+# extract width and height from a PNG file header inside a base64 string
+function png_wh(img::String)
+    # PNG header is 8 bytes, 4 byte chunk size, 4 byte IHDR string, 8 bytes for w, h
+    decoded = base64decode(img[1:32])  # Base64 encodes 6 bits per character
+    if any(decoded[13:16] .!= b"IHDR")  # check if the header looks reasonable
+        throw(ArgumentError("Base64-encoded PNG has a badly formed header."))
```
stevengj, line 98: I don't think we should throw an error here; we should let the front end decide what to do with malformed PNG data (i.e. just don't set any metadata).

stevengj:
I don't understand why this can't be done by the front-end display.   It seems weird to do this in the kernel, because the kernel doesn't know anything about the display, and indeed the same kernel may be talking to multiple displays in theory.

stevengj:
Is there a Jupyter/Jupyterlab issue for this as a front-end feature?

src/inline.jl
xzackli: Makes sense, I'll make this change.

xzackli:
> Is there a Jupyter/Jupyterlab issue for this as a front-end feature?

I've made an issue at https://github.com/jupyterlab/jupyterlab/issues/9440 to see what people think. Maybe it's historical.

tfiers:
A short alternative, for if you only care about retina'ing PyPlot Figures, and don't mind that matplotlib's `(figwidth, figheight)` is not the exact same as the actual png's.

```julia
using IJulia

function IJulia.metadata(x::PyPlot.Figure)
    w, h = (x.get_figwidth(), x.get_figheight()) .* x.get_dpi()
    return Dict("image/png" => Dict("width" => w/2, "height" => h/2))
end
```

DilumAluthge:

@xzackli If you're still interested in working on this PR, let me know, and I can re-open it.

PR #937, opened 2020-07-20
Support Cell Macros (similar to IPython cell magics but easily user-definable)

cduck:
Adds the ability to apply macros to the contents of an entire cell in IJulia.  The use case that prompted me to implement this can be seen [here](https://github.com/cduck/FastRepl.jl#ijulia).

Simple use case (assume soft scope is disabled):
```julia
### Cell 1 ###
using SoftGlobalScope
```
```julia
### Cell 2 ###
@@softscope
x=0
for i in 1:10 x+=i end
x
```
Equivalent to:
```julia
@softscope begin  # :toplevel block
    x=0
    for i in 1:10 x+=i end
    x
end
```

Features:
- Runs any number of macros listed at the beginning of a cell.  The lowest macro is run first.
- Expressions on the same line as a cell macro are included as arguments.  The cell contents are the last argument.
- Automatically runs the macros in IJulia.cell_macros afterward.  If soft scope is enabled, `var"@softscope"` is included here.
- Include `@@noauto` in the cell to disable the automatic macros for that cell.

I'm relatively new to Julia so please give me suggestions if my code style or macro usage could be improved.  I can add tests and documentation.

### DIFF
--- a/src/execute_request.jl
+++ b/src/execute_request.jl
@@ -16,7 +16,7 @@ import REPL: helpmode
 # use a global array to accumulate "payloads" for the execute_reply message
 const execute_payloads = Dict[]
 
-
+# An array of macros to run on the contents of each cell
 const cell_macros = []
 
 function run_cell_code(code)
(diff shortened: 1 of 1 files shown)


cduck:
Bump.

stevengj:
I'm not sure we want to encourage people to use syntax in Jupyter notebooks that can't be copied-and-pasted to other Julia code.  They can always use regular macros, string macros, etcetera…

DilumAluthge:

@cduck If you're still interested in working on this PR, let me know, and I can re-open it.

PR #944, opened 2020-08-14
TagBot: use DOCUMENTER_KEY

DilumAluthge:
This will ensure that, after the next release is registered, that the tag will correctly build the documentation.

Once the documentation has been built on a tag, the "stable" link will work.

cc: @stevengj

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -2,6 +2,7 @@ name: TagBot
 on:
   schedule:
     - cron: 0 0 * * *
+  workflow_dispatch:
 jobs:
   TagBot:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@stevengj Now that https://github.com/JuliaRegistries/General/pull/19534 is merged, you can manually trigger TagBot as follows:

Go to https://github.com/JuliaLang/IJulia.jl/actions?query=workflow%3ATagBot

It looks like this:

Screen Shot 2020-08-14 at 13 26 18

Click on the gray "Run workflow:" button:

Screen Shot 2020-08-14 at 13 27 00

Now click on the green "Run workflow" button.

TagBot will run and create a tag. This will automatically trigger Documenter to build the docs on the tag. Once those docs are build, the "stable" link should work.

---

Note: you don't have to run TagBot manually. TagBot will automatically run one time each day. However, the benefit of running TagBot manually this one time is that it will build the "stable" docs.

After this first time, you won't need to ever run TagBot manually.

stevengj:
Thanks, it all works now!

PR #945, opened 2020-08-15
Organize/add docstrings in API docs

jlumpe:
Replaced single `@autodocs` block with explicit references to all functions/types/constants with docstrings and organized them under headers (also tried to sort them between the Public API / internals pages as best I could). Added a bunch more things to `@docs` blocks in internals page which seem like they should be documented but do not currently have docstrings.

### DIFF
--- a/docs/src/library/internals.md
+++ b/docs/src/library/internals.md
@@ -1,6 +1,12 @@
 # Internals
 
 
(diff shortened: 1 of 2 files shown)


stevengj:
Thanks!

PR #947, opened 2020-08-20
support repl_ast_transforms in Julia 1.5

stevengj:
See also JuliaLang/julia#37047


stevengj:
Reverting this change since it seems to interact badly with Revise (#960), and in general it is not clear to me if transformations that are designed to work with the REPL will always work with IJulia.

PR #948, opened 2020-09-01
Add kwarg julia_cmd to installkernel (to support JupyterHub with containers, etc.)

oschulz:
This PR makes it possible to specify a custom `julia` command via

```julia
installkernel(Julia, julia_cmd="my_custom_julia_start_script")
```

It also allows passing an array - this should be very helpful to people (like me :-) ) who often run Julia in Linux containers but may be forced (e.g. by using a site-wide JupyterHub) to have Jupyter running outside of the container.

For Docker, running

```julia
installkernel(
    "Julia via Docker",
    julia_cmd = [
        "docker", "run", "--rm", "--net=host",
        "--volume=/home/USERNAME/.local/share/jupyter:/home/USERNAME/.local/share/jupyter",
        "some-container", "/opt/julia-1.x/bin/julia"
    ]
)
```

from within a container instance now works - provided Docker has been set up in a way that supports things like this (e.g. the user is  "USERNAME" both inside and outside). I use Docker in this example because it's well known, though it is not necessarily the right runtime for such use cases (see below).

More and more scientific computing sites now offer JupyterHub instances and also support containers via runtimes like [Shifter](https://github.com/NERSC/shifter) or [Singularity](https://sylabs.io/)). Like with Docker, calling Julia ([or Python, for that matter](https://docs.nersc.gov/services/jupyter/#shifter-kernels-on-jupyter)) becomes a multi-part command (only simpler), e.g. `["shifter", "--image", "myimage", "/opt/julia/bin/julia"]`. This typically forces users to manually fiddle with their "kernel.json" files, which is inconvenient.

Having a `julia_cmd` kwarg (supporting both a simple string and a vectors of strings) should make life easier.

### DIFF
--- a/deps/kspec.jl
+++ b/deps/kspec.jl
@@ -47,6 +47,7 @@ end
 
 """
     installkernel(name::AbstractString, options::AbstractString...;
+                  julia_cmd::Cmd,
                   specname::AbstractString,
                   env=Dict())
 
@@ -66,8 +67,23 @@ The returned `kernelpath` is the path of the installed kernel directory, somethi
 directory (which defaults to `name` with spaces replaced by hyphens).
 
 You can uninstall the kernel by calling `rm(kernelpath, recursive=true)`.
+
+You can specify a custom `julia` command via `julia_cmd`. For example,
+you may want specify that the Julia kernel is running in a Docker
+container (but Jupyter will run outside of it), by calling `installkernel`
+from within such a container instance like this (or similar):
+
+```
+installkernel(
+    "Julia via Docker",
+    julia_cmd = `docker run --rm --net=host
+        --volume=/home/USERNAME/.local/share/jupyter:/home/USERNAME/.local/share/jupyter
+        some-container /opt/julia-1.x/bin/julia`
+)
+```
 """
 function installkernel(name::AbstractString, julia_options::AbstractString...;
+                   julia_cmd::Cmd = `$(joinpath(Sys.BINDIR,exe("julia")))`,
                    specname::AbstractString = replace(lowercase(name), " "=>"-"),
                    env::Dict{<:AbstractString}=Dict{String,Any}())
     # Is IJulia being built from a debug build? If so, add "debug" to the description.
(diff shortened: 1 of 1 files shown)


fredrikekre:
Perhaps it would be nicer to give this as a `Cmd`? Or at least support that too. You can get out the string array with `c.exec`.
````
julia> c = ```
       docker run
           --rm
           --net=host
           --volume=/home/USERNAME/.local/share/jupyter:/home/USERNAME/.local/share/jupyter
           some-container
           /opt/julia-1.x/bin/julia
       ```
`docker run --rm --net=host --volume=/home/USERNAME/.local/share/jupyter:/home/USERNAME/.local/share/jupyter some-container /opt/julia-1.x/bin/julia`

julia> c.exec
7-element Array{String,1}:
 "docker"
 "run"
 "--rm"
 "--net=host"
 "--volume=/home/USERNAME/.local/share/jupyter:/home/USERNAME/.local/share/jupyter"
 "some-container"
 "/opt/julia-1.x/bin/julia"
````

oschulz:
> Perhaps it would be nicer to give this as a Cmd

I thought about that, but I thought it might look inconsistent with the `options` argument. But if there are no objections in that regard, using a `Cmd` would of course be more user-friendly.

oschulz:
If we use `Cmd`, the cleanest solution might be not to support `String` and `AbstractVector{<:AbstractString}` as well, but `Cmd` only.

What would you guys prefer?

stevengj:
I'm inclined to support `Cmd` only.

Also, probably just call the keyword argument `julia` rather than `julia_cmd`.

fredrikekre:
> I thought about that, but I thought it might look inconsistent with the options argument.

Yea, perhaps the options argument should be `Cmd` too.

oschulz:
@stevengj sounds good

oschulz:
> Yea, perhaps the options argument should be Cmd too.

Wouldn't that be abusing `Cmd` a bit, though? Not having an actual command in front, I mean?

oschulz:
Sorry, lost track of this for a bit. I've implemented the changes now, using `julia_cmd::Cmd`.

oschulz:
Ok, now tests pass, too. :-)

oschulz:
@fredrikekre, @stevengj , does this look good to you now?

fredrikekre (approved):
LGTM. Is it possible to add a test? Perhaps just run `installkernel` and check that it produces the output file you want?

oschulz:
@fredrikekre, @stevengj , gentle bump regarding merge.

oschulz:
Sorry for bumping this again, @stevengj, I know you guys are busy. I have some users who could really use this.

stevengj:
LGTM, but as I commented above I would rather call the keyword `julia` rather than `julia_cmd`.

oschulz:
> as I commented above I would rather call the keyword julia rather than julia_cmd.

oschulz:
> as I commented above I would rather call the keyword julia rather than julia_cmd.

Oops, sorry @stevengj, I overlooked your comment. It's changed to `julia` now.

oschulz:
Thanks!

PR #958, opened 2020-10-30
Allow users to opt out of the default kernel.

vchuravy:
Shared Jupyter installations might want to be more selective.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,8 +1,10 @@
 using Conda
 
-# Install Jupyter kernel-spec file.
-include("kspec.jl")
-kernelpath = installkernel("Julia", "--project=@.")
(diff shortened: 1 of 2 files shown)


stevengj:
Should it be stored in the `prefsfile` so that this setting is remembered on updates (which needs to get migrated to the new Pkg-based system)?   In which case it should be `IJULIA_INSTALLKERNEL=no` or ``IJULIA_INSTALLKERNEL=yes` (the default), maybe?

vchuravy:
For the use case I need this, it's easy to just set this in the script that brings up Julia + IJulia, and I am probably fine with users manually upgrading IJulia and installing a new kernelspec.

Preferences.jl only works on Julia 1.6.

stevengj:
Fine; can you add a note to the README?

stevengj:
I added a note to the documentation.

PR #962, opened 2020-11-18
fuzzy completions

aviatesk:
This PR provides a way to get fuzzy completions from IJulia kernel.

Fuzzy completions are useful in those frontends where some richer UIs to show candidates are available, and I think most of Jupyter frontends suit for that, e.g.:
image

Users can opt-in fuzzy completion mode by `installkernel(...; fuzzy_completion_mode = true)` and then [FuzzyCompletions.jl](https://github.com/JunoLab/FuzzyCompletions.jl) will provide completions as a drop-in replacement of the `REPL.REPLCompletions` module.


travigd:
Won't this cause weird issues because fuzzycompletions isn't in the project.toml?

aviatesk:
It is in Project.toml, no ?
https://github.com/JuliaLang/IJulia.jl/blob/a1166d8a151cf963acf48ab64ba48ce55eac85dc/Project.toml#L9

travigd:
Ah sorry, it is. I must have missed it.

I do wish there was a better way to do "optional" dependencies in Julia though

aviatesk:
well, just as for FuzzyCompletions, it's really light weight and so I don't think this dependency could be any problem.

stevengj:
It would be better to have a plug-in architecture for completions (ideally in the underlying REPL.REPLCompletions stdlib), so that users can just put `using FuzzyCompletions` in their startup file and it will use it.

MasonProtter:
Now that we have package extensions, and weak deps, could we revive this?

stevengj:
A package extension seems reasonable.

JamesWrigley:
Since it's such a light dependency I'm kinda inclined to just include it by default. Does the REPL support completion plugins yet? If not I'm inclined to go ahead with this since who knows how long it'll take to get REPL support.

PR #964, opened 2020-11-29
Create greet.yml to welcome first time contributor

logankilpatrick:


logankilpatrick:
We have something very similar working here: https://github.com/JuliaLang/www.julialang.org/blob/master/.github/workflows/greet.yml

fredrikekre:
It seems unnecessary to spend compute on this, in particular since (i) this repo has barely no traffic and (ii) you are already welcomed with this

> If you have a question please search or post to our Discourse site: https://discourse.julialang.org.
> We use the GitHub issue tracker for bug reports and feature requests only.
> 
> If you're submitting a bug report, be sure to include as much relevant information as
> possible, including a minimal reproducible example and the output of `versioninfo()`.
> If you're experiencing a problem with a particular package, open an issue on that
> package's repository instead.
> 
> Thanks for contributing to the Julia project!

logankilpatrick:
That's fair. It makes sense just for PR's then. Feel free to close this if you think that's the right course of action

tk3369:
I was just about to suggest that we make a massive set of PR’s to every registered packages 😝

Issue #968, opened 2020-12-06
IJulia kernel doesn't work for Julia 1.6 on macOS


crstnbr:
I can't get the IJulia kernel to work when using Julia 1.6/master. Tried latest IJulia release and IJulia#master. The kernel either seems to not start all or die immediately.

Julia <= 1.5.3 kernels work just fine.

```Julia
julia> versioninfo()
Julia Version 1.6.0-DEV.1661
Commit 56dd7d71a7* (2020-12-04 13:18 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.6.0)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = subl
  JULIA_NUM_THREADS = 6
  JULIA_PROJECT = /Users/crstnbr/.julia/environments/base
  JULIA_DQMC = /Users/crstnbr/sciebo/codes/julia-sdw-dqmc
  JULIA_PKG_SERVER = pkg.julialang.org
```

![kernel_issues](https://user-images.githubusercontent.com/187980/101287330-ea25f680-37ef-11eb-87af-251a841374b4.gif)

stevengj:
Does the IJulia build succeed?  Mine is failing due to https://github.com/JuliaLang/Pkg.jl/issues/2270

crstnbr:
The build succeeds.

![ijulia 2020-12-06 22_03_54](https://user-images.githubusercontent.com/187980/101292310-11d88700-380f-11eb-8abb-c23a93c56584.gif)

stevengj:
Can you try turning on debugging output?  https://julialang.github.io/IJulia.jl/stable/manual/troubleshooting/#Debugging-IJulia-problems

crstnbr:
Hm, I can't manage to get any useful debugging output. It seems like setting `ENV["IJULIA_DEBUG"]=true` doesn't do anything?

I removed all instances of IJulia (in `~/.julia/packages` and `~/.julia/compiled`) and used a fresh environment to `add` and `build` IJulia with the environmental variable set. When using `IJulia.notebook()` to start the jupyter server within the REPL I only get this output with no information whatsoever:

[...]

PR #976, opened 2021-01-18
MacOS fixes for issue #968

rgobbel:
Changes to run with Julia >1.5 on MacOS: heartbeat calls removed, dynamically-created
redirect functions replaced by one statically-created function to get around
problems with incremental compilation.

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -23,7 +23,9 @@ const publish = Ref{Socket}()
 const raw_input = Ref{Socket}()
 const requests = Ref{Socket}()
 const control = Ref{Socket}()
-const heartbeat = Ref{Socket}()
+@static if VERSION < v"1.6" || !Sys.isapple()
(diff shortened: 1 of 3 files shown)


travigd:
I don't think that removing the heartbeat socket is an option. Jupyter uses unresponsive heartbeats to detect dead kernels.

rgobbel:
> I don't think that removing the heartbeat socket is an option. Jupyter uses unresponsive heartbeats to detect dead kernels.

Well, that's unfortunate. The segfault is still there if I put the heartbeat back in, so something deeper is broken. Even *with* the heartbeat, Julia kernels appear to hang around forever if not killed by hand via the Shutdown command in the notebook. I just tried it with a fresh install on 1.5.3, after making sure there were no cached version of any IJulia code hanging around. Quitting the server from the shell did not shut down the Julia kernel.

Issue #977, opened 2021-01-19
Julia processes are not automatically shut down on Jupyter server shutdown


rgobbel:
When closing an IJulia notebook, unless you do an explicit "shutdown kernel" command from within the notebook, its associated Julia process continues to run after the Jupyter server has terminated. Verified with Julia 1.5.3 and 1.6.0-beta1, on MacOS Catalina and Ubuntu 20.04.

TheCedarPrince:
As @rgobbel reported, I too have this issue on Fedora 32. I am running Julia 1.5.0 and IJulia version 1.23.1. Thanks for the help!

TheCedarPrince:
Seems like this issue is related as well to bug #921. Any updates on this @stevengj or @fredrikekre?

stevengj:
Seems like this should be reported to the Jupyter repo, not IJulia?

rgobbel:
> Seems like this should be reported to the Jupyter repo, not IJulia?

I thought this was specifically an IJulia problem, but as far as I can tell everything works fine with Jupyter Lab--clean shutdown and no dangling processes whether I use the "Quit" command or just type a ctrl-C.

ruizhi92:
I encountered similar problems on a linux cluster. A lot of julia process still exist even after complete shut down of Julia and Jupyter-lab. Any updates on this?

gwd999:
same behavior even more "openly visible" if you use `jupyter console --kernel=julia` 
in that case `exit()` will not even close the typical IJulia input zou can only help yourself via `CTRL-Z` but a `ps` will cintinue to show you the `jupyter-console` process which not even a `kill` command will be able to get rid off

PR #978, opened 2021-01-20
Fix segfault on macos (#968): doing heart beat via async and ZMQ.jl

### LINKED ISSUE #968: IJulia kernel doesn't work for Julia 1.6 on macOS

daviehh:
May fix #968. Instead of running the heart beat in a thread via `uv_thread_create`, just do the echo server in a new process.


rgobbel:
This versions runs, though on my Macbook Pro running Catalina one Julia process keeps running after shutdown via control-C at the command line. It does shut down properly if done via the Shutdown Kernel command in the notebook.

daviehh:
@rgobbel Thanks for confirming it runs.

As you noted in https://github.com/JuliaLang/IJulia.jl/issues/977, the lingering process with ctrl-c may a separate issue; 
this PR just fixes the ijulia not working with julia 1.6.0, issue #968 

Whereas, the #977 issue may be a different thing:
- quitting from jupyter from Shutdown Kernel command in the notebook and the process quits normally
- quitting by ctrl-c in terminal: by turning debugging on, it turns out both processes (the main ijulia process and the worker heartbeat process) actually correctly quits, but for some reason jupyter then decides to restart the kernel. you may notice that the pid of the julia processes is different from before the ctrl-c shutdown. Looks like the shutdown_reply did not get sent in
 https://github.com/JuliaLang/IJulia.jl/blob/2300ab9cc4c723f9fabec5ea9b0360870b898494/src/handlers.jl#L186-L189

stevengj:
Won't using `addprocs` like this interfere with people doing multiprocessing?

daviehh:
@stevengj thanks, good call! Well, since the echo server blocks, things like `@everywhere` or `pmap` would work, but it would interfere with e.g. hard-coded `@spawnat`, etc, and would expose the process in e.g. `workers()`

Alternatively, the new process could be launched with e.g. `@async run(`Base.julia_cmd() -e [...]`)`; and it could be terminated by sending the echo server a string other than "ping" (which is currently the heartbeat message)
[link](https://github.com/jupyter/jupyter_client/blob/dd89e2748129e457d01cfe38ff79ba65d835fa9e/jupyter_client/channels.py#L148-L151)

e.g. in the julia hearbeat server
```julia
while true
    msg = recv(sock, String)
    println(msg)
    if msg == "ping"
        send(sock, msg)
    else
        exit()
    end
end
```

one could then send e.g. "stop" to the channel upon quitting IJulia (either in the `shutdown_request` [function](https://github.com/JuliaLang/IJulia.jl/blob/2300ab9cc4c723f9fabec5ea9b0360870b898494/src/handlers.jl#L185) or registering an `atexit` [hook](https://docs.julialang.org/en/v1/base/base/#Base.atexit))

stevengj:
Note also that the whole point of the heartbeat thread, as I understand it, is so that Jupyter can detect if the process dies, so handling it in a separate process seems like it defeats the point?

daviehh:
Thanks for pointing that out, the latest commit now uses async to call send/recv from ZMQ.jl instead of uv_thread_create and not in a new process.

stevengj:
async also won't work, because if you have a long-running computation in Julia that doesn't `yield` then Jupyter will think Julia has died and restart it.

PR #985, opened 2021-02-12
fix heartbeat on 1.6

### LINKED ISSUE #968: IJulia kernel doesn't work for Julia 1.6 on macOS

stevengj:
Fixes #968, closes #978.

The issue seems to be that, in Julia 1.6, the first time you do a `ccall` it calls `dlsym` at runtime and updates some global state, which is not thread-safe in a non-Julia thread (whereas in 1.5 this was done at compile-time).  Doing the `dlsym` outside the thread seems to fix the problem.

cc @rgobbel, @staticfloat


staticfloat (approved):

crstnbr:
Maybe worth tagging a new release?

PR #987, opened 2021-02-19
Set version to 1.23.2.

fredrikekre:
Looks to be only bugfixes since last release so 1.23.2 seems appropriate.


fredrikekre:
#989

Issue #986, opened 2021-02-13
redirect_stdio is broken by Julia 1.7


stevengj:
As described in https://github.com/JuliaLang/IJulia.jl/issues/968#issuecomment-762459535 and https://github.com/JuliaLang/IJulia.jl/issues/968#issuecomment-776272479, the `stdio` redirection code will need to change for Julia 1.7 to use the new mechanism from https://github.com/JuliaLang/julia/pull/39132 — currently, starting the IJulia kernel in Julia 1.7 gives an error:
```
ERROR: LoadError: ArgumentError: expecting stdout stream
Stacktrace:
 [1] (::Base.redirect_stdio)(io::IJulia.IJuliaStdio{Base.PipeEndpoint})
   @ IJulia ~/.julia/dev/IJulia/src/stdio.jl:30
 [2] init(args::Vector{String})
   @ IJulia ~/.julia/dev/IJulia/src/init.jl:109
 [3] top-level scope
   @ ~/.julia/dev/IJulia/src/kernel.jl:24
in expression starting at /Users/stevenj/.julia/dev/IJulia/src/kernel.jl:24
```

stevengj:
(Seems to work fine in Julia 1.6.)

PR #989, opened 2021-02-22
Proposed fix for #986.

rgobbel:
No metaprogramming required now. Fix for #986. No changes were required other than deleting these few lines.


src/stdio.jl
```diff
@@ -23,16 +23,6 @@ Base.displaysize(io::IJuliaStdio) = displaysize(io.io)
 Base.unwrapcontext(io::IJuliaStdio) = Base.unwrapcontext(io.io)
 Base.setup_stdio(io::IJuliaStdio, readable::Bool) = Base.setup_stdio(io.io.io, readable)
 
-for s in ("stdout", "stderr", "stdin")
```
fredrikekre, line 26: You should put this in a conditional to still support older Julia version:
```
if VERSION < v"1.7.0-DEV.254"
    # ...
end
```

rgobbel: Ah, of course. Done, testing now...

src/stdio.jl
```diff
@@ -23,13 +23,15 @@ Base.displaysize(io::IJuliaStdio) = displaysize(io.io)
...
+if VERSION < v"1.7.0-DEV.254"
+    for s in ("stdout", "stderr", "stdin")
+        f = Symbol("redirect_", s)
+        sq = QuoteNode(Symbol(s))
+        @eval function Base.$f(io::IJuliaStdio)
+            io[:jupyter_stream] != $s && throw(ArgumentError(string("expecting ", $s, " stream")))
+            Core.eval(Base, Expr(:(=), $sq, io))
+            return io
+        end
     end
```
stevengj, line 35: Shouldn't there be an `else` clause to do something new in Julia 1.7?

rgobbel: There's no need. This code added functionality that is now part of the standard versions of these functions.

PR #1007, opened 2021-06-10
Note about default environment

lamorton:
If you add IJulia to a different environment, then Jupyter won't be able to launch the IJulia kernel automatically.  I got bit by this recent.

### DIFF
--- a/docs/src/manual/installation.md
+++ b/docs/src/manual/installation.md
@@ -5,7 +5,7 @@
 
 First, [download Julia](http://julialang.org/downloads/) *version 0.7
 or later* and run the installer.  Then run the Julia application
-(double-click on it); a window with a `julia>` prompt will appear.  At
+(double-click on it); a window with a `julia>` prompt will appear.  After ensuring that you have activated the default Julia environment, at
 the prompt, type:
 ```julia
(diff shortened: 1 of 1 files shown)


fredrikekre:
I always run IJulia from a non-default env. Can you clarify what didn't work for you @lamorton ?

lamorton:
@fredrikekre If you follow the instructions `using Pkg; Pkg.add("IJulia")` when you are not in the default environment, then `IJulia` only gets added to your non-default environment.  When Jupyter tries to launch a Julia kernel, it evidently uses the default Julia environment, where `IJulia` is still not available.  Jupyter gets stuck with a message like "Unable to connect to Julia kernel."

fredrikekre:
>  When Jupyter tries to launch a Julia kernel, it evidently uses the default Julia environment

Are you launching jupyter from outside of Julia then?

lamorton:
@fredrikekre Yes, typically Jupyterhub.

fredrikekre:
Okay makes sense. If you launch it through Julia this isn't a problem (e.g. `IJulia.notebook(; dir=pwd())` and similarly for jupyterhub)

Issue #1012, opened 2021-06-23
Code sharing with VS Code Jupyter Notebook implementation


davidanthoff:
We are getting closer and closer to finishing the native Jupyter Notebook support in VS Code for Julia (see [here](https://github.com/julia-vscode/julia-vscode/pull/980)). When I started, I copied a few functions from IJulia.jl over, but as things progressed I ended up copying a fair bit of code and now there is probably too much code duplication.

Should we try to arrange this is a way that instead of the VS Code extension copying code we move the shared code into its own package, say JupyterCore.jl, that both IJulia.jl and the Julia extension for VS Code can use? Probably mostly a question for @stevengj, whether he would be on board with that. This is not super urgent, I think we could start out by shipping the extension with all the code duplication and then also clean this up down the road.

For background, the notebook implementation in VS Code will _not_ use the kernelspec configuration that IJulia.jl installs nor the jupyter protocol. Instead, it will be a "no configuration" solution, i.e. if a user has the Jupyter and Julia extension for VS Code installed, everything will work without any further things. The whole execution aspect of things is handled by the native notebook execution API that VS Code started shipping recently. In the VS Code Julia extension we can't take a full dependency on IJulia.jl because we in general cannot take dependencies on packages that pull in anything binary (so in this case MbedTLS.jl and maybe Conda.jl), that wouldn't work with the way we deploy Julia packages in the extension.

stevengj:
So their notebook API doesn't use ZMQ, but uses the same JSON message protocol over some other transport layer?

I'm fine with an IJuliaCore.jl package if that makes sense…

davidanthoff:
VS Code now has a [notebook API](https://code.visualstudio.com/updates/v1_57#_notebook-api) that is an in-process JavaScript API, like all the other APIs that VS Code provides. VS Code extensions can just directly interface with that, and the pending PR on the Julia extension does that. At that level we (the Julia extension) essentially tell VS Code that we can execute Julia code in notebooks and provide a callback that does so.

[...]

PR #1016, opened 2021-07-27
Move things to IJuliaCore.jl

davidanthoff:
Fixes https://github.com/JuliaLang/IJulia.jl/issues/1012.

This is still very WIP, i.e. I still need to fix tests and make sure that everything actually works. The corresponding package is https://github.com/JuliaLang/IJuliaCore.jl, which is not yet registered (but already used by the Julia VS Code extension).

@stevengj I'll finish this, i.e. clean things up more, get tests to pass etc. But if you have a moment to look over this and just generally give feedback whether you are OK with this rough direction that would be great, before I start to polish the details.


stevengj:
Looks reasonable to me… not sure if anything is happening here?

PR #1028, opened 2021-12-02
fix `jupyter kernelspec` invalid name warning

xuestrange:
Kernel names can only contain ASCII letters and numbers and these separators: - . _ (hyphen, period, and underscore)

if the name contains other symbols like `( )` brackets and space, jupyter will give a warning and substitute all illegal symbol to `-`
image


fakerms:
Visual name can have blank in it, so it is better to change the default logic for converting name to specname.

PR #1004 has done this.

Issue #1021, opened 2021-08-17
IJulia in local environment, dead kernel


arifyunando:
Current Outcome:
Recently, I freshly installed Julia 1.6.2 in my computer. I made a dedicated environment for a project with Pkg.generate("envName") along with Python venv. After activating python venv and installing jupyter with pip, I open Julia and set environment variable as below:
* ENV["PYTHON"] = Sys.which("python")
* ENV["JUPYTER"] = Sys.which("jupyter") 

Those 2 commands set the environment variable to the binaries in the venv folder. After activating environments with Pkg.activate("envName"), I proceed with installing IJulia with Pkg.add("IJulia") within generated TOML file. The installation went normally without any error, and when I run IJulia.notebook(dir="path/to/project"), jupyter notebook opened just fine. However, when I open a notebook with IJulia kernel, it tries to start the kernel for a while and then throws "Dead Kernel" prompts to me.

Expected Outcome : 
The kernel started successfully and I can use Julia in my notebook.

Action taken : 
I tried updating IJulia and build it again [Fail], I tried installkernel("Julia", "--project="path/to/project.toml") [Fail]. I also tried to open jupyter notebook directly from cmd (not Julia REPL), and it throws me this message
```cmd
ERROR: LoadError: ArgumentError: Package IJulia not found in current path:
- Run `import Pkg; Pkg.add("IJulia")` to install the IJulia package.

Stacktrace:
 [1] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:893
in expression starting at C:\Users\arify\.julia\packages\IJulia\e8kqU\src\kernel.jl:1
[I 11:32:40.564 NotebookApp] KernelRestarter: restarting kernel (1/5), new random ports
ERROR: LoadError: ArgumentError: Package IJulia not found in current path:
- Run `import Pkg; Pkg.add("IJulia")` to install the IJulia package.

Stacktrace:
 [1] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:893
in expression starting at C:\Users\arify\.julia\packages\IJulia\e8kqU\src\kernel.jl:1
[I 11:32:43.580 NotebookApp] KernelRestarter: restarting kernel (2/5), new random ports
ERROR: LoadError: ArgumentError: Package IJulia not found in current path:
- Run `import Pkg; Pkg.add("IJulia")` to install the IJulia package.

Stacktrace:
 [1] require(into::Module, mod::Symbol)
[...]

Issue #1002, opened 2021-04-26
IJulia not working on Julia 1.61. on Windows 10


xiaodaigh:
I have searched the issues and I found a similar issue marked as solve in 2018 for Julia 0.7. But it's not working for me. I can see the Julia kernel but when I start a notebook I get 

![image](https://user-images.githubusercontent.com/4497189/116027577-99c50d80-a698-11eb-915a-f0bbe7702be8.png)

![image](https://user-images.githubusercontent.com/4497189/116027582-a0538500-a698-11eb-87f8-16d90b8d6f79.png)

I have tried to `]build IJulia` and reinstall Conda.jl. The python kernel works. Just the Julia one doesn't work. I have deleted the whole `.julia` folder and it still doesn't work. 

```
Julia Version 1.6.1
Commit 6aaedecc44 (2021-04-23 05:59 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = sublime_text
  JULIA_NUM_THREADS = 6
  JULIA_PKG_DEVDIR = c:/git/

````

**Comment by SimonDanisch** [comment]
I've been trying IJulia ~ 3-4 times per year for a couple of years on different platforms, and on windows on different machines, and I always run into this...
Is there anything we can do here?
I think IJulia hasn't been working for 2-3 years for me...On Linux I had a bunch of other issues...
Did I somehow manage to consistently have a badly setup environment, is this a widespread issue and IJulia isn't used that much anymore, or is everyone just going through a lot to get things going and I'm just too lazy to debug this?

**Comment by stevengj** [comment]
Does the Python kernel work?  (If so, it's unrelated to IJulia.)

**Comment by stevengj** [comment]
It works fine for me, but I don't use Windows.

Have you tried the troubleshooting tips?  e.g. building IJulia in debug mode to see if you get additional output? https://julialang.github.io/IJulia.jl/stable/manual/troubleshooting/#Debugging-IJulia-problems

**Comment by SimonDanisch** [comment]
```
julia> ENV["IJULIA_DEBUG"]=true
julia> using Pkg
julia> Pkg.build("IJulia")
[...]

Issue #1001, opened 2021-04-19
Environment of a new notebook


yha:
It seems like creating a new notebook (File->New Notebook) starts it in the same environment as the last open notebook. I'm not sure if this is deliberate, but I find it a bit surprising & confusing (especially since there is no visual indication of the environment in the interface).
I would expect a new notebook to be in the default environment, or perhaps in the same environment that `notebook()` was called from.
ref #820

travigd:
IJulia starts notebooks in the environment of the directory where the project and manifest files are (or if they don't exist, in the default environment).

If you have two notebooks in the same directory, they will use the same environment. To circumvent this, just use multiple directories (you have to anyway to create different project and manifest files).

yha:
> IJulia starts notebooks in the environment of the directory where the project and manifest files are

I'm not sure what that means. Which project and manifest files? 
My understanding is that if you open an existing notebook, it looks in its directory (where the notebook file is) and its superdirectories, which is fine. 
My issue is with *new* notebooks that were not saved yet (so there is no notebook file). These seem to use the directory/environment of whatever notebook was last open, as far as I can tell.

travigd:
I think that was a typo -- I think I meant to say 
> Julia starts notebooks in the environment of the directory where the notebook is (or if the project and manifest files don't exist, in the default environment).

Sorry for that.

> My issue is with new notebooks that were not saved yet (so there is no notebook file). These seem to use the directory/environment of whatever notebook was last open, as far as I can tell.

This is probably a JupyterLab behavior. The notebook files always exist (as far as I know) even if you haven't saved them yet (e.g., creating a new notebook will create a file on disk called "Untitled.ipynb" or whatever).

At the end of the day, IJulia is actually not even aware of the notebook files. From IJulia's (and IPython, IRKernel, etc.) perspective, it is started in a particular working directory and then just runs a server that speaks the Jupyter protocol in order to execute code (and the Jupyter protocol is just a fancy REPL - it has no idea about your notebook files).

If creating a new notebook starts it in the same directory as the previous notebook, that's just because Jupyter is spawning the Julia process in that directory. If that's the case, then this would be an upstream issue (with Jupyter Notebook or Jupyter Lab).

thompsonmj:
[...]

Issue #975, opened 2021-01-18
macOS: Julia unresponsive unless iJulia is in the global environment?


dpo:
Is it expected that IJulia must be installed in the global environment? I followed the troubleshooting steps in the documentation and confirmed that my firewall isn't the issue. If IJulia is installed in an environment other than the global environment, evaluating a cell never completes (the status remains  `[*]`).

fredrikekre:
I usually have it in my local environment so something else must be up. https://github.com/JuliaLang/IJulia.jl/issues/968 maybe if you are using Julia 1.6.

dpo:
I'm on Julia 1.5.3. I checked that a Python notebook works (and it does).

nilshg:
This can be closed.

PR #1030, opened 2021-12-17
Add notes on fixing "Connecting to kernel" failures.

jd-foster:
This adds to the docs on the issue of fixing "Connecting to kernel" failures. This questions seems to arise often among a variety of users and may be worth providing guidance on how to address it.

For reference, here is a list of times where this issue has arisen, played a role or could be helped by a better understanding here:
https://github.com/JuliaLang/IJulia.jl/issues/975
https://github.com/JuliaLang/IJulia.jl/issues/1001
https://github.com/JuliaLang/IJulia.jl/issues/1002
https://github.com/JuliaLang/IJulia.jl/issues/1021

and on Discourse:
https://discourse.julialang.org/t/ijulia-kernel-not-starting-unless-ijulia-is-installed-in-the-global-environment-ie-v1-1/21172/4
https://discourse.julialang.org/t/launch-jupyter-kernel-with-e-using-somepackage-flag/47980


stevengj:
Thanks.  Should those issues be closed now?

Issue #1035, opened 2022-03-17
support for pdf plot output


jjallaire:
Hi there, we've been putting some work into integrating support for IJulia into Quarto (<https://quarto.org>), which is a scientific publishing system based on Pandoc (sort of like R Markdown, in fact built by the same team, but fundamentally cross-language and cross-engine).

We have plots working correctly w/ various output formats, but one noteworthy limitation is that when creating a PDF the plot output is provided as PNG (as SVG currently doesn't work out of the box w/ LaTeX). As you can imagine this results in dramatically lower quality than is possible with native PDF plot output. One of Quarto's main uses cases is creation of scientific publications for academic journals, so getting high quality output ends up being pretty important.

I see the available output formats in IJulia enumerated here:

https://github.com/JuliaLang/IJulia.jl/blob/cc2a9bf61a2515596b177339f9a3514de8c38573/src/display.jl#L19-L28

Further, I noted the place where JuliaPlots is yielding output here:

https://github.com/JuliaPlots/Plots.jl/blob/master/src/ijulia.jl#L33-L56

It would be fantastic if these could be updated to produce PDF output. This might not at first blush seem to make sense for a  "notebook" but if you think of Jupyter as an execution engine for realizing executable code into output it makes more sense (i.e. think about `nbconvert` as the client rather than the traditional browser front end). This does in fact currently work for Matplotlib output on the Python side.

It's certainly a fine stopgap to use PNG output for plots in PDFs, but my guess is that when users start creating articles they want to submit for publication that the PDF support will quickly go from nice-to-have to must have.

jjallaire:
Once this lands we still might need to change Quarto to stop "redirecting" pdf to png here: https://github.com/quarto-dev/quarto-cli/blob/main/src/resources/jupyter/lang/julia/setup.jl#L22-L27

Further, we might want some sort of version check to make sure that we use the workaround when updated IJulia isn't avaialble.

jeremiahpslewis:
Ooh, good that you mention a workaround for earlier versions of IJulia (once/if PR is merged). I wanted to suggest a slightly different approach: what about a Quarto.jl package which bundles the logic which is currently in the setup.jl file, requires a couple of packages (Plots, IJulia), and sets a lower bound on the packages based on PDF support (and other things as they may come up in the future).

jjallaire:
[...]

PR #1038, opened 2022-04-29
Fix PDF render

jeremiahpslewis:
Fixes https://github.com/JuliaLang/IJulia.jl/issues/1035 alongside 

PR [#4188](https://github.com/JuliaPlots/Plots.jl/pull/4188) (which has been merged)

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -25,6 +25,7 @@ const ijulia_mime_types = Vector{Union{MIME, AbstractVector{MIME}}}([
         MIME("text/html"),
     ],
     MIME("text/latex"),
+    MIME("application/pdf"),
 ])
 
 """


jeremiahpslewis:
@fredrikekre Wanted to flag this as ready for review. :)

stevengj:
Is display of PDF images well supported in Jupyter?

This issue https://github.com/jupyter/notebook/issues/3652 seems to indicate that it only works in certain browsers or with certain plug-ins installed.

jeremiahpslewis:
I think you're right in pointing out there is a bit of a file type conflict here:

Per the issue you linked, browser support of PDFs is (potentially) incomplete.

But Latex workflows like quarto's use case which use Jupyter to build a document (but not display it in the browser) do not work well with SVGs and require the use of the PDF format for exporting / displaying vector graphics.

Right now, if you create a PDF document with R, Python, and Julia graphs via quarto, the Julia graph gets dumped out as a PNG and looks significantly pixelated, the other two languages (afaik) allow for PDF viz export and certainly lead to vector viz output in the PDF.

travigd:
The Jupyter format defines cell outputs as a dict that contains multiple representations of an output (e.g., plot). The frontend (e.g., jupyterlab) chooses which of the formats its want to render (so only one is *shown* at a time, but the various formats are present in the notebook file).

IJulia tries to generate a representation for every MIME type it supports (which is what this PR is changing).

This is actually not the case for Plots.jl -- it actually does override `IJulia.display_dict`: https://github.com/JuliaPlots/Plots.jl/blob/b6d7bd82e6c029b8cbbafa756015670cb0ab0013/src/init.jl#L92, which is what (personally) I think Quarto should do.

jeremiahpslewis:
@travigd Thanks for the nudge to investigate this further, the Plots.jl PR https://github.com/JuliaPlots/Plots.jl/pull/4188 solved the issue for Plots, so this isn't necessary.

And for Makie, turns out it's a one-liner which can be done on the Quarto side: `IJulia.register_mime(MIME("application/pdf"))` 🎉

benz0li:
> This is actually not the case for Plots.jl

Is there a way to override this in `startup.jl`/`startup_ijulia.jl`? For

1.  `Python`, one may set `c.InlineBackend.figure_formats = {'svg', 'pdf'}` in `ipython_config.py`/`ipython_kernel_config.py`
1.  `R`, one may set `options(jupyter.plot_mimetypes = c('text/plain', 'image/svg+xml', 'application/pdf'))` in `Rprofile.site`

in order to output both `image/svg+xml` and `application/pdf` plot formats in the notebook file:

*  `image/svg+xml` for plot display in the browser
*  `application/pdf` for direct PDF conversion

ℹ️ With `application/pdf` plot format in the notebook file, `inkscape` is not required for PDF conversion using `nbconvert`.

Issue #1043, opened 2022-06-23
Unlike in REPL, passing (e, catch_backtrace()) to the @error macro displays unreadable raw stacktrace instead of text representation


BoZenKhaa:
Following code in REPL displays the error message passed through `(e, catch_backtrace())` to the `@error` macro correctly:

```julia
julia> try
           a=b
       catch e
           @error "Oh no, exception!" (e, catch_backtrace())
       end
┌ Error: Oh no, exception!
│   (e, catch_backtrace()) =
│    UndefVarError: b not defined
│    Stacktrace:
│      [1] top-level scope
│        @ REPL[8]:2
│      [2] eval
│        @ ./boot.jl:373 [inlined]
│      [3] eval_user_input(ast::Any, backend::REPL.REPLBackend)
│        @ REPL ~/bin/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:150
│      [4] repl_backend_loop(backend::REPL.REPLBackend)
│        @ REPL ~/bin/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:246
│      [5] start_repl_backend(backend::REPL.REPLBackend, consumer::Any)
│        @ REPL ~/bin/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:231
│      [6] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool)
│        @ REPL ~/bin/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:364
│      [7] run_repl(repl::REPL.AbstractREPL, consumer::Any)
│        @ REPL ~/bin/julia-1.7.2/share/julia/stdlib/v1.7/REPL/src/REPL.jl:351
│      [8] (::Base.var"#930#932"{Bool, Bool, Bool})(REPL::Module)
│        @ Base ./client.jl:394
│      [9] #invokelatest#2
│        @ ./essentials.jl:716 [inlined]
│     [10] invokelatest
│        @ ./essentials.jl:714 [inlined]
│     [11] run_main_repl(interactive::Bool, quiet::Bool, banner::Bool, history_file::Bool, color_set::Bool)
│        @ Base ./client.jl:379
│     [12] exec_options(opts::Base.JLOptions)
│        @ Base ./client.jl:309
│     [13] _start()
│        @ Base ./client.jl:495
└ @ Main REPL[8]:4
[...]

Issue #1003, opened 2021-04-26
Stack traces on 1.6 don't use colors


marius311:
Julia 1.6-style stack trace coloring is really nice, but does not appear to be displayed in IJulia (1.23.2). However, it seems it could be since this is not a Jupyter limitation, i.e. stderr can use colors:

![image](https://user-images.githubusercontent.com/1510968/116154295-59b45980-a69d-11eb-99ba-a6f3b4d4648a.png)

Would be great to have stack traces use colors as well.

PR #1044, opened 2022-06-23
Replace SimpleLogger with ConsoleLogger

BoZenKhaa:
Closes https://github.com/JuliaLang/IJulia.jl/issues/1043 

Provides pretty printing of error messages in Jupyter.


BoZenKhaa:
Maybe does away with https://github.com/JuliaLang/IJulia.jl/issues/1003 as well?

PR #1053, opened 2022-09-01
Add Julia svg logo

steff456:
Hi!

This PR adds the Julia logo in svg format in the resources folder and it is related to https://github.com/jupyterlab/jupyterlab/issues/12997 where we want to fix the blurry icon present when a lot of zoom is made in jupyterlab.

The asset was downloaded from the official assets of Julia project available [here.](https://github.com/JuliaLang/julia-logo-graphics/blob/master/images/julia-dots.svg)


stevengj:
Thanks!

PR #1061, opened 2022-12-02
Warn against deleting `.julia` folder with deved packages.

baggepinnen:
The previous advice is potentially dangerous, this PR adds a warning for users who have called `pkg> develop` on any package


fredrikekre:
I would not even mentioning it as an option but instead recommend something more targeted. `.julia/packages/IJulia` probably, and perhaps `.julia/conda`?

stevengj:
Yes, I would just remove that advice entirely, and replace it with just removing `.julia/conda`.

Nowadays the package system is more robust than in the early days of IJulia, and it's pretty rare that you would need to remove `.julia` entirely.

es38989496:
不错

ViralBShah:
Update based on @stevengj's suggestion and merge?

docs/src/manual/troubleshooting.md
```diff
@@ -7,7 +7,7 @@
 problem you can type `Pkg.build()` to try to rerun the install scripts.
 * If you tried it a while ago, try running `Pkg.update()` and try again:
   this will fetch the latest versions of the Julia packages in case
-  the problem you saw was fixed.  Run `Pkg.build("IJulia")` if your Julia version may have changed.  If this doesn't work, you could try just deleting the whole `.julia` directory in your home directory (on Windows, it is called `Users\USERNAME\.julia` in your home directory) via `rm(Pkg.dir(),recursive=true)` in Julia and re-adding the packages.
+  the problem you saw was fixed.  Run `Pkg.build("IJulia")` if your Julia version may have changed.  If this doesn't work, you could try just deleting the whole `.julia` directory in your home directory (on Windows, it is called `Users\USERNAME\.julia` in your home directory) via `rm(Pkg.dir(),recursive=true)` in Julia and re-adding the packages. Before you delete the `.julia` directory, make sure you do not have unsaved work in any repository beloinging to developed packages in the subfolder `.julia/dev`
```
stevengj, line 10:
```suggestion
  the problem you saw was fixed.  Run `Pkg.build("IJulia")` if your Julia version may have changed.  If this doesn't work, you could try just deleting the whole `.julia/conda` directory in your home directory (on Windows, it is called `Users\USERNAME\.julia\conda` in your home directory) via `rm(abspath(first(DEPOT_PATH), "conda"),recursive=true)` in Julia and re-adding the packages. 
```

stevengj: Update to recommend deleting just ~/.julia/conda

Issue #1091, opened 2023-08-16
Kernel exception: usage_request


marius311:
On some particular cluster's Jupyterlab, I keep geting these messages periodically in random cells:

```
KERNEL EXCEPTION
KeyError: key "usage_request" not found

Stacktrace:
 [1] getindex(h::Dict{String, Function}, key::String)
   @ Base ./dict.jl:484
 [2] eventloop(socket::ZMQ.Socket)
   @ IJulia ~/.julia/packages/IJulia/6TIq1/src/eventloop.jl:8
 [3] (::IJulia.var"#14#17")()
   @ IJulia ./task.jl:514
```

<details>

Julia installed from binaries, IJulia v1.24.0, Jupyterlab 3.6.5

```
Julia Version 1.9.2
Commit e4ee485e909 (2023-07-05 09:39 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 128 × AMD EPYC 7763 64-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver3)
  Threads: 129 on 128 virtual cores
Environment:
[...]

PR #1092, opened 2023-08-18
ignore unknown request types

stevengj:
Jupyter extensions can generate arbitrary new request types, apparently, which this PR ignores (rather than throwing an exception).  Workaround for #1091.


marius311:
Friendly ping if this could make it into a release at some point?

PR #1114, opened 2024-05-20
Prefer to `show` `AbstractString`s with text mimes

### LINKED ISSUE #1113: Broken display of `AbstractString`s when preferred MIME type `istextmime`
### Description

Users implementing `show` methods for an `AbstractString` will have a broken display if the preferred MIME type `istextmime`. A motivating example is in Typstry.jl, which implements `show` for `"image/png"` and `"image/svg+xml"`. Both of these methods work in Pluto.jl, but only `"image/png"` works in IJulia.jl. This is because `"image/svg+xml"` is preferred over `"image/png"`, but this format assumes that an `AbstractString` is ["raw data"](https://github.com/JuliaLang/IJulia.jl/blob/185533665803d5c0f9de6103d0c386422d9e1c6a/src/inline.jl#L24). [Instead of dispatching to `show`, it is returned as a `String`](https://github.com/JuliaLang/IJulia.jl/blob/185533665803d5c0f9de6103d0c386422d9e1c6a/src/inline.jl#L41-L43).

I don't really understand this assumption, why is it being used? The only thing I have found so far is the same assumption in [`repr`](https://docs.julialang.org/en/v1/base/io-network/#Base.repr-Tuple{MIME,%20Any}) and presume that's where it came from. A notable difference is that users are able to implement `repr` to patch in a fix.

### Possible Solutions

- Remove the assumption the an `AbstractString` is "raw data"
    - I don't know if or what relies on this assumption
- Export `israwtext`
    - Users can implement this to override the assumption
    - Requires packages to depend on IJulia.jl
- Change [this line](https://github.com/JuliaLang/IJulia.jl/blob/185533665803d5c0f9de6103d0c386422d9e1c6a/src/inline.jl#L41) to `String(repr(mime, x; context = InlineIOContext(buf)))`
    - Users can implement a fix for `repr`, while preserving the preexisting behavior otherwise because `repr` uses the same assumption
    - Extra work to convert some values to a `String`

### 1. The output of `versioninfo()`

IJulia v0.24.2

```
julia> versioninfo()
Julia Version 1.10.3
Commit 0b4590a5507 (2024-04-30 10:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, tigerlake)
Threads: 8 default, 0 interactive, 4 GC (on 8 virtual cores)
Environment:
  JULIA_EDITOR = hx
```

### 2. How you installed Julia

juliaup

### 3. A minimal working example (MWE), also known as a minimum reproducible example

```julia
julia> using IJulia

julia> struct S <: AbstractString end

julia> Base.ncodeunits(::S) = 0;

julia> Base.iterate(::S) = nothing;

julia> Base.show(io::IO, ::MIME"image/svg+xml", ::S) = print(io, "s");

julia> S()
""

julia> show(stdout, "image/svg+xml", S())
s
julia> IJulia.display_dict(S())
Dict{String, Union{String, JSON.Writer.JSONText}} with 2 entries:
  "text/plain"    => "\"\""
  "image/svg+xml" => ""
```

jakobjpeters:
`israwtext` assumed that an `AbstractString` with a text MIME was already in the requested format. Instead, only assume this if the string is not otherwise `showable`. This enables customizing the display of `AbstractString`s.

Closes #1113.

### DIFF
--- a/src/inline.jl
+++ b/src/inline.jl
@@ -22,7 +22,7 @@ const ipy_mime = [
 # need special handling for showing a string as a textmime
 # type, since in that case the string is assumed to be
 # raw data unless it is text/plain
-israwtext(::MIME, x::AbstractString) = true
+israwtext(m::MIME, x::AbstractString) = !showable(m, x)
 israwtext(::MIME"text/plain", x::AbstractString) = false
 israwtext(::MIME, x) = false
 


stevengj:
LGTM.

PR #1135, opened 2024-12-10
Refactor heartbeat to shutdown cleanly

halleysfifthinc:
With Julia 1.11.2 (also tested 1.10.7), I'm seeing segfaults when shutting down an idle
Julia kernel (with current IJulia@v1.26) in Jupyter on Linux.
```
% jupyter --version
Selected Jupyter core packages...
IPython          : 8.30.0
ipykernel        : 6.29.5
ipywidgets       : 8.1.5
jupyter_client   : 8.6.3
jupyter_core     : 5.7.2
jupyter_server   : 2.14.2
jupyterlab       : 4.3.2
nbclient         : 0.10.1
nbconvert        : 7.16.4
nbformat         : 5.10.4
notebook         : 7.0.7
qtconsole        : not installed
traitlets        : 5.14.3
```

~~Reproducible by opening/starting a notebook, and then shutting down after the kernel is
connected and idle.~~ EDIT: I can no longer reproduce this, but the change still seems like a good idea?

There are several errors in the segfault dump, but the heartbeat thread is notable

The ZMQ docs state "zmq_proxy() runs in the current thread and returns only
if/when the current context is closed."

The heartbeat socket doesn't need to be global, as nothing else touches
it. So, if we create the heartbeat socket in a `Context` that has a global ref,
we can close the context, which will cause zmq_proxy to return and then
that thread to end/finish.

With this change, I'm not sure whether the `jl_gc_safe_enter` ccall is necessary?


src/heartbeat.jl
```diff
@@ -19,14 +22,20 @@ function heartbeat_thread(sock::Ptr{Cvoid})
...
-    ccall(zmq_proxy[], Cint, (Ptr{Cvoid}, Ptr{Cvoid}, Ptr{Cvoid}),
-          sock, sock, C_NULL)
-    nothing
+    ret = ZMQ.lib.zmq_proxy(heartbeat, heartbeat, C_NULL)
+    @static if VERSION ≥ v"1.9.0-DEV.1588" # julia#46609
+        # julia automatically "adopts" this thread because
+        # we entered a Julia cfunction.  We then have to enable
+        # a GC "safe" region to prevent us from grabbing the
+        # GC lock with the call to zmq_proxy, which never returns.
+        # (see julia#47196)
```
stevengj, line 31:
```suggestion
        # leave safe region if zmq_proxy returns (when context is closed)
```

src/heartbeat.jl
```diff
@@ -7,10 +7,13 @@
...
 const threadid = zeros(Int, 128) # sizeof(uv_thread_t) <= 8 on Linux, OSX, Win
-const zmq_proxy = Ref(C_NULL)
+const zmq_proxy_context = Ref{Context}()
 
 # entry point for new thread
-function heartbeat_thread(sock::Ptr{Cvoid})
+function heartbeat_thread(heartbeat_addr::Cstring)
+    zmq_proxy_context[] = Context()
+    heartbeat = Socket(zmq_proxy_context[], ROUTER)
+    GC.@preserve heartbeat_addr bind(heartbeat, unsafe_string(heartbeat_addr))
```
stevengj, line 16: I'd rather create the context and the socket in `start_heartbeat`, and simply pass the `heartbeat` socket pointer to `heartbeat_thread`.

halleysfifthinc: The ZMQ docs recommend creating the sockets inside the thread they are used from, and that was doable since the heartbeat socket isn't used anywhere else. But, I don't think there is any advantage to that vs. creating in `init.jl` like it was originally.

halleysfifthinc:
Bump

JamesWrigley (approved):
LGTM! Entering the GC safe region is still necessary since the heartbeat thread behaves much the same as it did before.

I would like to replace the heartbeat with a task running `send(sock, recv(sock))` in a loop, but then the heartbeat would be blocked by anything blocking from the client if Julia was started with only one thread. We could also run that send/recv pair within `heartbeat_thread()` but then we'd have to ensure it doesn't throw exceptions or return anything non-C supported which also feels awkward...

Note that technically this doesn't guarantee that the heartbeat thread has stopped since we don't explicitly wait for it to finish, but in practice it's probably overwhelmingly likely since we send a network request and `sleep` before exiting.

src/init.jl
```diff
@@ -87,7 +88,8 @@ function init(args)
     raw_input[] = Socket(ROUTER)
     requests[] = Socket(ROUTER)
     control[] = Socket(ROUTER)
-    heartbeat[] = Socket(ROUTER)
+    heartbeat_context[] = Context()
+    heartbeat = Socket(heartbeat_context[], ROUTER)
```
JamesWrigley, line 92: @halleysfifthinc, should this be indexing like `heartbeat[] = ...` since `heartbeat` is still a Ref?

halleysfifthinc: D'oh, bad rebase on my part. Yes that should be indexed.

PR #1138, opened 2024-12-19
Jupyter messaging updates/correctness

### LINKED ISSUE #953: IJulia not generating standardised metadata for cells within jupyter lab to show cell execution duration time
```
version() output
```
image

As can be seen from the above image, the jupyterlab-execute-time extension only shows the datetime stamp when using IJulia kernel but the execution duration is missing. Works fine for python / spark / sparkr / pyspark kernels. Having the same issue for IRkernel ticket was opened there also.

Have already validated the issue with at least one other user using a different version of OS/kernel than mine and have also reached out to the jupyterlab-execute-time devs who have confirmed that this likely appears to be an issue with the kernel and not with jupyterlab or the jupyterlab-execute-time plugin.

This issue seems to be related to the metadata that gets generated for each cell by the kernel.
image

Below you can find the cell metadata generated using Python kernel for example:
image

Please see the comments here - https://github.com/deshaw/jupyterlab-execute-time/issues/14

### LINKED ISSUE #1000: comm_info_reply should not be sent on IOPub
IJulia sends the `comm_info_reply` message (in response to `comm_info_request`) on the IOPub channel, but it should be sent as a reply on the shell channel.

https://jupyter-client.readthedocs.io/en/stable/messaging.html#comm-info

halleysfifthinc:
- Bump Julia compat to be able to use `pkgversion` in `"kernel_info_request"` to report the IJulia version
- Some messages and responses were sent on the wrong channel:
  - Replies should be sent to the channel the request was sent on
    - This is not :100: explicit in the messaging protocol docs, but appears to be the norm. Some messages are only described as always being sent (by the frontend/client) on a specific channel, but e.g. I saw `"kernel_info_request"` sent on both the shell/requests channel *AND* the control channel. (The latter is not described as possible according to the docs.)
- Replies should [always](https://jupyter-client.readthedocs.io/en/latest/messaging.html#request-reply) have a `{ 'status': 'ok' }`
- The `transient` metadata field was added in protocol version 5.1. Without the ability to add transient metadata, `"update_display_data"` messages can't be used to update/redisplay existing outputs (while participating in/with the rest of the IJulia display machinery). Related PR #946 is more extensive and stalled; this is the minimum necessary change to support this.

Fixes #953
Fixes #1000

Reference: https://jupyter-client.readthedocs.io/en/latest/messaging.html


halleysfifthinc:
Bump

src/IJulia.jl
```diff
@@ -36,7 +36,7 @@ export notebook, jupyterlab, installkernel
 using ZMQ, JSON, SoftGlobalScope
 import Base.invokelatest
 import Dates
-using Dates: now
+using Dates: now, format, @dateformat_str, UTC, ISODateTimeFormat
```
JamesWrigley, line 39: Is `@dateformat_str` needed?

JamesWrigley (changes requested):
Not strictly necessary for this PR, but we should start using `jupyter_kernel_test` in the tests: https://github.com/jupyter/jupyter_kernel_test

src/IJulia.jl
halleysfifthinc: Ah! No. 
(In an earlier version of this PR I was manually recreating the `ISODateTimeFormat` with the `'Z'` at the end, but that is no longer needed by the current code.)

JamesWrigley (approved):
LGTM!

PR #1140, opened 2025-01-14
Make message receive and handling async

### LINKED ISSUE #858: Async comm stuff
Will fill this out later. :^)

Want to be able to handle comm message asynchronously (e.g. allow manipulating an slider while something is running).

Ping @stevengj @shashi

halleysfifthinc:
## Motivation

All messages from the front-end/server are received and handled synchronously, including [custom comm messages](https://jupyter-client.readthedocs.io/en/latest/messaging.html#custom-messages) (`comm_open`, `comm_msg`, and `comm_close`). So, any currently executing cell blocks the IJulia kernel from receiving and handling any IOPub/comm messages. For example, in the following WebIO MWE, a JS function updates an "output" `Observable`, and the JS function is triggered by setting an ("input") observable:

```julia
using WebIO, Observables
s = Scope()
s["in"] = Observable{String}("")
s["out"] = Observable{String}("")
onjs(s["in"], js"""
function (val)
    _webIOScope.setObservableValue("out",val);
end""")
```

you can't observe a new `s["out"]` value (aka the result of the JS function) during execution of the same cell that set `s["in"]` (which triggers the JS function).

<details><summary>Example Julia function that fails (hangs) without async comms</summary>
<p>

```julia
function julia_js_julia(_in, out, str)
    ch = Channel{String}()
    obsf = on(out) do val
        put!(ch, val)
    end
    t = @async take!(ch)
    _in[] = str
    out = fetch(t)
    off(obsf)
        
    return out
end
```

*This example function isn't thread-safe. (The `scp["in"]` observable isn't locked, so concurrently setting it could lead to interleaved/mismatched updates to the `scp["out"]` observable.)
</p>
</details> 

One example of an actual use-case/benefit is `PlotlyJS.to_image`, which uses the same 
Julia => JS => Julia observable setup to retrieve the results of a plotly.js function call.
Currently, the `PlotlyJS.to_image` function soft-fails because the observable that holds the generated
image is only updated after the current cell finishes execution (when IJulia can process the
`comm_msg` from WebIO in the Jupyter frontend/client).

## Testing

I've manually tested that the above WebIO MWE works with this PR, and that interrupting still works. I realize this is a fairly fundamental rearchitecturing of the message receiving/handling, but I'm not sure what else to test and/or if there is a good way to test any of this in CI. I'm open to any hints/pointers if you want more thorough testing/test cases.

Fixes #858.

P.S. Breadcrumb for the future: This new architecture has a lot of parallels (easily adapted) to the new [subshells feature](https://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-subshells) that was recently implemented in ipython/ipykernel#1249.

### DIFF
--- a/src/eventloop.jl
+++ b/src/eventloop.jl
@@ -1,74 +1,122 @@
 """
-    eventloop(socket, kernel)
+    eventloop(socket, kernel, msgs, handlers)
 
 Generic event loop for one of the [kernel
 sockets](https://jupyter-client.readthedocs.io/en/latest/messaging.html#introduction).
 """
-function eventloop(socket, kernel)
-    task_local_storage(:IJulia_task, "write task")
-    while true
+function eventloop(socket, kernel, msgs, handlers)
+    while isopen(msgs)
         try
-            local msg
-            try
-                msg = recv_ipython(socket, kernel)
-            catch e
-                if isa(e, EOFError)
-                    # The socket was closed
-                    return
-                else
-                    rethrow()
+            while isopen(msgs)
+                msg = take!(msgs) # can throw if `msgs` is closed while waiting on it
+                try
+                    send_status("busy", kernel, msg)
+                    invokelatest(get(handlers, msg.header["msg_type"], unknown_request), socket, kernel, msg)
+                catch e
+                    if e isa InterruptException && kernel.shutting_down[]
+                        # If we're shutting down, just return immediately
+                        return
+                    elseif !isa(e, InterruptException)
+                        # Try to keep going if we get an exception, but
+                        # send the exception traceback to the front-ends.
+                        # (Ignore SIGINT since this may just be a user-requested
+                        #  kernel interruption to interrupt long calculations.)
+                        content = error_content(e, msg="KERNEL EXCEPTION")
+                        map(s -> println(orig_stderr[], s), content["traceback"])
+                        send_ipython(kernel.publish[], kernel, msg_pub(kernel.execute_msg, "error", content))
+                    end
+                finally
+                    flush_all()
+                    send_status("idle", kernel, msg)
                 end
-            end
-
-            try
-                send_status("busy", kernel, msg)
-                msg_type = msg.header["msg_type"]::String
-                invokelatest(get(handlers, msg_type, unknown_request), socket, kernel, msg)
-            catch e
-                if e isa InterruptException && kernel.shutting_down[]
-                    # If we're shutting down, just return immediately
-                    return
-                elseif !isa(e, InterruptException)
-                    # Try to keep going if we get an exception, but
-                    # send the exception traceback to the front-ends.
-                    # (Ignore SIGINT since this may just be a user-requested
-                    #  kernel interruption to interrupt long calculations.)
-                    content = error_content(e, msg="KERNEL EXCEPTION")
-                    map(s -> println(orig_stderr[], s), content["traceback"])
-                    send_ipython(kernel.publish[], kernel, msg_pub(kernel.execute_msg, "error", content))
-                end
-            finally
-                flush_all()
-                send_status("idle", kernel, msg)
+                yield()
             end
         catch e
-            if kernel.shutting_down[]
+            if kernel.shutting_down[] || isa(e, ZMQ.StateError) || isa(e, InvalidStateException)
+                # a ZMQ.StateError is almost certainly because of a closed socket
+                # an InvalidStateException is because of a closed channel
                 return
-            end
-
-            # the Jupyter manager may send us a SIGINT if the user
-            # chooses to interrupt the kernel; don't crash on this
-            if isa(e, InterruptException)
-                continue
-            elseif isa(e, ZMQ.StateError)
-                # This is almost certainly because of a closed socket
-                return
-            else
+            elseif !isa(e, InterruptException)
+                # the Jupyter manager may send us a SIGINT if the user
+                # chooses to interrupt the kernel; don't crash for that
                 rethrow()
             end
         end
+        yield()
     end
 end
 
 """
     waitloop(kernel)
 
-Main loop of a kernel. Runs the event loops for the control and shell sockets
+Main loop of a kernel. Runs the event loops for the control, shell, and iopub sockets
 (note: in IJulia the shell socket is called `requests`).
 """
 function waitloop(kernel)
-    control_task = @async eventloop(kernel.control[], kernel)
-    kernel.requests_task[] = @async eventloop(kernel.requests[], kernel)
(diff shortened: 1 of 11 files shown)


JamesWrigley:
This sounds like a good idea, but it absolutely needs tests before merging. At some point I'll start writing tests for more of the internals which you should be able to modify for this PR, but feel free to have a go already if you have time :)

halleysfifthinc:
:+1: I will wait until you've added more internals tests before I do anything further. I am/have been running IJulia with this PR to give any bugs the opportunity to surface.

JamesWrigley:
If you rebase this on master I think we can continue with it :slightly_smiling_face: Couple things:
- We should use `Threads.@spawn` instead of `@async`.
- We should run CI with multiple threads by default to try to catch any race conditions.

JamesWrigley:
Keeping it on the interactive threads make sense, but for that we should use `Threads.@spawn :interactive`. `@async` has the unfortunate side-effect of pinning the parent task to the same thread so it's kinda discouraged now.

halleysfifthinc:
> pinning the parent task to the same thread

Right.. Is that not equivalent to `Threads.@spawn :interactive`? The rest of the IJulia kernel is synchronous/not using tasks, so it will always be on the first (aka interactive) thread, and we want the rest of the IJulia activity to stay on that thread too, just allowed to be asynchronous/concurrent?

Happy to learn more if I'm wrong, this was my first serious foray into async/concurrent programming!

JamesWrigley:
That is technically true, but `@async` is still deprecated so I'd prefer we stick with `Threads.@spawn` and explicitly specifying the threadpool. One other advantage is that if there's multiple threads in the interactive threadpool then we can use all of them instead of one.

src/eventloop.jl
```diff
@@ -1,73 +1,122 @@
...
-        end
-    catch e
-        if IJulia._shutting_down[]
-            return
-        end
-
-        # the Jupyter manager may send us a SIGINT if the user
-        # chooses to interrupt the kernel; don't crash on this
-        if isa(e, InterruptException)
-            eventloop(socket, kernel)
```
halleysfifthinc, line 52: I can't think of a reason for this to be recursive instead of the outer loop as I reorganized it too?

halleysfifthinc (commented):
I've left some comments to explain some design decisions and/or about open questions I have.

I'm still unsure how to add tests for this, and I'd welcome any brainstorming.

JamesWrigley:
Sorry I missed this :see_no_evil: I'll try to review it this week but feel free to ping me if I forget.

JamesWrigley (commented):
I'm not quite convinced that what we're doing here is safe. If I understand correctly the reasoning is:
1. ZMQ sockets are not thread-safe.
2. Thus we use `@async` to ensure that all tasks are running on the same thread.
3. Thus we can safely recv/send in different tasks as long as we lock appropriately to prevent one recv being interleaved with another recv (likewise for send)

But that's making the assumption that ZMQ.jl's `recv` and `send` don't do anything to the socket internally that may conflict with each other, and I don't think that's true. Imagine this sequence:
- Task 1 is sending and yields immediately after calling `zmq_msg_send()`: https://github.com/JuliaInterop/ZMQ.jl/blob/1e1b458180311b19127937e8dd0befa79a93d54f/src/comm.jl#L8
  Let's say that `zmq_msg_send()` fails because we have to try again (EAGAIN).
- Task 2 is receiving and yields immediately after calling `zmq_msg_recv()`: https://github.com/JuliaInterop/ZMQ.jl/blob/1e1b458180311b19127937e8dd0befa79a93d54f/src/comm.jl#L80
  Let's say it fails for some non-EAGAIN reason (maybe a corrupted message or something). This overwrites the internal error code from `zmq_msg_send()`.
- Control switches back to Task 1 which calls `zmq_errno()`, which returns the error code from the call to `zmq_msg_recv()` and thus incorrectly fails instead of trying again.

Now I'm pretty sure that neither `send()` or `recv()` will yield in those places so in practice this particular situation couldn't happen right now, but that's an implementation detail of ZMQ and certainly not something we can rely on. But I also can't think of a good alternative yet 🤔

Also, I fixed some lingering-task issues in #1190 which seems to have caused some merge conflicts, sorry about that 🙈

halleysfifthinc:
I'm stumped for the moment, so I pushed the current state of things for a second opinion on any obvious design flaws.

Summary of design constraints as I understand them:
- Sending by the kernel must be synchronous (implicitly assumed by required ordering of e.g. status messages, etc), but receiving can be asynchronous.
- Lack of thread-safety in these ZMQ sockets prevents interleaved send/recv actions, therefore a single (thread-safe) lock must be used to limit send/recv actions to only one thread/task at a time.
- The original (synchronous) `eventloop` had a default/"resting" state of holding the socket locks while waiting to receive. This is untenable for an asynchronous `eventloop` because a receiver task would immediately re-acquire the socket lock after a successful read, preventing any other tasks from reliably/fairly acquiring the lock to send.

The solution to avoiding a receiver task's default/"resting" state of holding the lock is obviously polling, so that the lock is only held when we can immediately read without blocking, quickly followed by a release. I had originally looked into polling, but couldn't figure how to get ZMQ polling to interact with Julia's task scheduler properly.

Using the new ZMQ.Poller, the current design of receiving tasks is basically
```julia
poller = ZMQ.Poller([kernel.requests[]])
while true
    pr = wait(poller)
    msg::Msg = recv_ipython(pr.socket, kernel)
    # route via Channel to other async eventloop task ...
end
```

while other tasks handle the messages, where handling may require sending/receiving on at least 3 different sockets (an execute_request that reads from stdin would send and/or receive from the IOPub, stdin, and Shell sockets).

> The recv socket will only ever be used from the conductor task so only the send socket needs a lock in case multiple tasks try to send stuff.
Agreed, and unless I'm missing something, I'm not actually sure why a second `inproc` socket is needed at all with my current design. By keeping a shared send/recv lock per socket, only one task can use the socket at any given time, and no task should be able to be blocked while holding the lock.

Unfortunately, the kernel hangs during the precompile script after line 53 (sending the "complete_request"). Some `println`s at the beginning and end of the locked regions suggest that its not deadlocked by the socket locks, and a healthy sprinkling of `println`s in the receiver tasks and precompile script resulted in getting just past line 55 (receiving the response). Any ideas?

JamesWrigley:
Turned out to be a bug in the Poller, an internal flag was not being reset properly: https://github.com/JuliaInterop/ZMQ.jl/pull/266

Now precompilation works for me with 6b435e4, but the kernel tests are hanging somewhere :thinking: Could you have a look at where? Lemme know if it looks like another poller bug and I'll investigate.

PR #1141, opened 2025-01-28
Add coverage to CI

JamesWrigley:
Also updated the `julia-actions/cache` version.


JamesWrigley:
Oof, that is an unfortunate percentage :sweat_smile: https://app.codecov.io/github/JuliaLang/IJulia.jl/tree/coverage
I'll have a go at adding more tests in the future.

PR #1142, opened 2025-01-28
Documentation fixes and a changelog

JamesWrigley:
I originally just wanted to add a changelog but ended up fixing other docs things too :sweat_smile: In particular:
- Moved some files into the `docs/` directory to play nicely with LiveServer.jl (a5c7377)
- Add missing docstrings to fix Documenter warnings (3441177)

And added a changelog in f6b8243. Since the new docstrings are for internal functions it would probably be good for someone who's more familiar with them than me to check that they're correct-ish.


JamesWrigley:
(if there's no objections, I'll merge this next week)

stevengj:
Thanks!

PR #1143, opened 2025-01-30
Add Aqua to the test suite

JamesWrigley:
Turns out we commit type piracy on `Base.display()`, so the piracy test is disabled for now until we figure out if it's needed. @stevengj, do you know if it's needed? AFAICT the method isn't called anywhere in the code base and it's not documented either.

Also removed some unnecessary dependencies in 4885721.


JamesWrigley:
(I'll go ahead and merge this, we can figure out the pirated method later)

Issue #1074, opened 2023-05-03
Add precompile statements to reduce startup latency


MasonProtter:
So @timholy and I have done a little preliminary looking into adding `precompile` statements to reduce the latency of IJulia kernels, and it seems to us that a major problem is that there's no clear way to run a PrecompileTools.jl or SnoopCompile.jl workflow to actually get the relevant precompile statements from a running notebook, so we're forced to resort to `--trace-compile` kernels. 

@stevengj  do you have any insight into how we might be able to set up a PrecompileTools.jl workflow that starts up and spins down a kernel during precompilation?

stevengj:
Maybe just copy https://github.com/JuliaLang/IJulia.jl/blob/cc2a9bf61a2515596b177339f9a3514de8c38573/src/kernel.jl but comment out the final `IJulia.waitloop()` line?   Then run `julia kernelcopy.jl`.

JamesWrigley:
I also would like to improve the TTFX. Some random thoughts (as someone who only recently started looking at the code):
- A lot of the codebase is not tested, which I'm aiming to improve in the future. Having something run in a test suite has a lot in common with running in a precompile workload so I think once the code is testable it'll be easier to work with under precompilation.
- There's a *lot* of global variables with pointers. Those all need to be cleaned up after the workload so they aren't serialized or we'll run into things like this: https://github.com/JuliaInterop/ZMQ.jl/issues/241. It may be best to bundle all of them into a `Kernel` struct that can be `close`'d (which would make testing easier).
- `waitloop()` is the thing we really want to precompile and the only way to 'cancel' it is by throwing an exception to the task or sending a `shutdown_request` to the kernel (which has the side effect of exiting the process so we probably shouldn't use that). I think it should be refactored to allow a graceful cancellation. This would make testing easier and it would make it easier to run in a precompilation workload.
- Not sure if the separate heartbeat thread will cause problems.

@MasonProtter, would you by any chance have time to look into some of these things with me? I will try to find some time for it over the next few weeks/months.

MasonProtter:
Yes I would be interested in collaborating on this!

PR #1145, opened 2025-02-05
Giant refactor to move all state into a Kernel struct

JamesWrigley:
Pretty much what the title says :stuck_out_tongue: This is not ready yet but I'm opening it for transparency. There are two reasons for the refactor:
1. Easier testing, so that we can more confidently make changes like #1140 and #1138.
2. Allow for creating a temporary kernel that can be used in a precompilation workload for #1074.

Overview:
- Almost all global state has been moved into the new `Kernel` struct, and all relevant _internal_ methods now require a kernel be passed to them. That's the bulk of the diff. I've tried to minimize other changes as much as possible, so there should :crossed_fingers: be no change in behaviour from `master`.
- There are certain fields like `ans` and `n` which are documented as globals and must stay global, so `setproperty!(::Kernel,...)` is implemented such that it will copy changes to those fields to their corresponding global variable. Similarly there are certain public methods like `IJulia.history()` that must still work without a kernel argument, so a new `_default_kernel` global is defined for normal usage and that's used by default for all public methods.

  An edge-case are the dicts `In` and `Out` which are by default assigned to the corresponding fields of `_default_kernel` and will not contain changes from temporary test kernels. I don't think that's a particularly important discrepancy so I'm inclined to ignore it.
- I implemented a basic test in `waitloop.jl` that spawns the IJulia waitloop and then makes a request to it using `jupyter_client`. This isn't working very well right now because blocking calls to the client causes hangs, I suspect there's a deadlock somewhere.

This will need a lot more work before it can be merged, but feel free to review the design already (cc @stevengj, @fredrikekre).

@MasonProtter, you should be able to do... something... with this for a precompilation workload. I would suggest creating a `Kernel`, connecting to some of the sockets with ZMQ.jl, and then sending some mock requests following the wire protocol: https://jupyter-client.readthedocs.io/en/latest/messaging.html#the-wire-protocol


JamesWrigley:
Figured out the deadlock, we can now start writing full tests with `jupyter_client` :)

src/IJulia.jl
```diff
@@ -33,13 +33,16 @@ The `IJulia` module is used in three ways
 module IJulia
 export notebook, jupyterlab, installkernel
 
+import MbedTLS
 using ZMQ, JSON, SoftGlobalScope
-import Base.invokelatest
+import Base: invokelatest, RefValue
```
stevengj, line 38: Probably `using` rather than `import` since we only use these, not extend them?   Also `invokelatest` is exported automatically these days.

(This code dates back to the days before `using Foo: bar`)

src/IJulia.jl
```diff
@@ -49,11 +52,140 @@ const depfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
...
+    # These fields are isbits types and need to be assigned explicitly to their
+    # global counterparts.
+    if name ∈ (:ans, :n, :In, :Out, :inited, :verbose)
+        setproperty!(IJulia, name, x)
+    end
+
+    setfield!(value, name, x)
+end
+
+function Base.close(kernel::Kernel)
```
stevengj, line 145: should be called by a `finalizer`?

src/IJulia.jl
```diff
@@ -49,11 +52,140 @@ const depfile = joinpath(dirname(@__FILE__), "..", "deps", "deps.jl")
...
+    # queue of objects to display at end of cell execution
+    displayqueue::Vector{Any} = Any[]
+end
+
+function Base.setproperty!(value::Kernel, name::Symbol, x)
+    # These fields are isbits types and need to be assigned explicitly to their
+    # global counterparts.
+    if name ∈ (:ans, :n, :In, :Out, :inited, :verbose)
+        setproperty!(IJulia, name, x)
+    end
```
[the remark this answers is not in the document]
stevengj: whereas `n` and `verbose` at least can probably be moved in to the `Kernel`?

src/stdio.jl
```diff
@@ -274,8 +260,8 @@ function oslibuv_flush()
 end
 
 import Base.flush
-function flush(io::IJuliaStdio)
+function flush(io::IJuliaStdio, kernel)
```
stevengj, line 263: This is a user-visible function, e.g. if the user calls `flush(stdout)`.

It would be better to store a reference to the kernel in the `IJuliaStdio` object.

src/IJulia.jl
JamesWrigley: Agreed, but I think I'd prefer to do those sorts of changes in a later PR. This one is already quite large :sweat_smile:

src/IJulia.jl
JamesWrigley: Unfortunately that wouldn't be easy because it `wait`'s on the stdio tasks so it might yield. I think it's ok to rely on the do-constructor closing it in the tests.

src/IJulia.jl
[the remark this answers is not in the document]
JamesWrigley: You're quite right :thinking: I need to figure out a good way to test that, for now I've added a `@test_broken` test in d9acb6e.

> whereas `n` and `verbose` at least can probably be moved in to the `Kernel`?

`n` is documented to be module-global (https://julialang.github.io/IJulia.jl/dev/library/public/#IJulia.n) so I don't think we can move that fully into `Kernel`, but `verbose` should be possible. I think only `@vprintln`/`@verror_show` will need tweaking to look at `_default_kernel.verbose` instead of the global.

src/stdio.jl
JamesWrigley: Good idea, fixed in d9acb6e.

JamesWrigley:
Progress update:
- Moved the waitloop task into `Kernel` to make testing easier, and added a `wait(::Kernel)` method to wait for it to finish.
- Added/expanded tests for autocompletion, MIME display, and `clear_output()` using `jupyter_kernel_test`. I think that covers everything we can use `jupyter_kernel_test` for.
- Moved the calls to `watch_stdio()` and `pushdisplay()` into `init()`, now `close(::Kernel)` will ensure they're cleaned up properly.
- Added a bunch of tests for some of the public API: `clear_history()`, `load()`, the hook system, and `history()`.
- Tweaked `history()` to print all the entries on new lines in 20dc591. From the docstring, I think that's what it should be doing?

To-do:
- Address remaining review comments.
- Add a test for `kernel.jl` to ensure that the kernel works outside of the test suite.
- Figure out why the `load()` test is failing on Windows.
- Go through the coverage reports to find more low-hanging fruit to test.

src/IJulia.jl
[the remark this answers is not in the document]
JamesWrigley: Fixed the tests for `ans`/`In`/`Out` being propagated to `kernel.current_module` in 2fd607f, also moved `verbose` fully into `Kernel`.

JamesWrigley:
I believe this is ready to be reviewed now. One thorny issue is segfaults on windows that appear to be coming from ZMQ, using `@threadcall` for the heartbeat seemed to fix it: d491800
But I'm not convinced I understand the problem fully, will try to debug it a bit more before merging.

For reviewers, I've tried to keep the commits atomic so I would recommend reviewing each commit one-by-one.

JamesWrigley:
Explicitly waiting for the heartbeat thread also seemed to fix it, which seems preferable to `@threadcall` in the interest of minimizing changes. But now mac is hanging...

JamesWrigley:
Back in the green with 0b6e7be :sweat_smile: I think this is in pretty good shape now.

JamesWrigley:
I went ahead and added a simple precompilation workload in 9bb03a1. On my system it reduces the running time of getting the kernel info and executing `42` from 4s to 0.5s.

Not sure why the tests are failing, it seems to be related to CondaPkg :thinking: Will investigate later, but I believe this is ready for review anyway.

JamesWrigley:
Back in the green, had to explicitly specify `MicroMamba` after the latest CondaPkg update.

MasonProtter:
Testing locally and installing a kernel, the kernel seems to die as soon as I attempt to connect

JamesWrigley:
Is the path to `kernel.jl` in the `kernel.json` file correct? If precompilation succeeds but Jupyter can't connect to the kernel I think that points to a problem in `kernel.json` (I ran into that while testing locally).

src/eventloop.jl
```diff
@@ -1,17 +1,28 @@
...
+                    # The socket was closed
+                    return
+                else
+                    rethrow()
+                end
+            end
+
+            try
+                send_status("busy", kernel, msg)
+                invokelatest(get(handlers, msg.header["msg_type"], unknown_request), socket, kernel, msg)
```
MasonProtter, line 25: This appears to hit `comm_open`, `comm_msg` and `comm_close`, but those don't have the same signature as the one used here

src/comm_manager.jl
```diff
@@ -97,7 +97,7 @@ end
 
 # handlers for incoming comm_* messages
 
-function comm_open(sock, msg)
+function comm_open(sock, msg, kernel)
```
MasonProtter, line 100:
```suggestion
function comm_open(sock, kernel, msg)
```
needed to match the signature from the event loop (I think?)

MasonProtter (commented):
Figured out the problem I was having, it was crashing at startup for me because `comm_open`, `comm_close`, and `comm_msg` had the wrong signature relative to the `invokelatest` call in the event loop.

Github isn't letting me give a suggestion for the `comm_close` and `comm_msg` functions, but they need to all have the same signature.

JamesWrigley:
Ouch, that's quite a serious oversight :grimacing: I've added tests and hopefully fixed the comms code in bb7cef4, with the one change of making the `:notarget` `Comm` sent in the `comm_open` error case to not be a primary comm, so it won't send a `comm_open` message to the client: https://github.com/JuliaLang/IJulia.jl/blob/4390ef4eca084464a932f8110d7c8d7c12ef5471/src/comm_manager.jl#L114

src/eventloop.jl
JamesWrigley: Fixed in bb7cef4.

stevengj:
@shashi, can you common on the `close_comm` question?

JamesWrigley:
Rebased to fix merge conflicts. About `close_comm()`, in the absence of clarification I'd suggest we go ahead with the existing behaviour for now since it's been working so far.

JamesWrigley:
(bump)

JamesWrigley:
(rebased to fix merge conflicts from #1149)

JamesWrigley:
(bump)

halleysfifthinc (approved):

MasonProtter:
IS it known why Windows is failing to precompile here? It seems to build and run fine on my machine.

JamesWrigley:
Nope, I still need to debug that (bit swamped this week). But glad it works on your machine now :tada: Do you see a TTFX improvement from the precompilation?

MasonProtter:
Yes, I am seeing a significant reduction in TTFX!

MasonProtter:
I should mention though that my main usecase (https://github.com/emacs-jupyter/jupyter) remains super slow the first time and seems to be unaffected by this. But baby steps, and I don't think that should hold back this effort.

JamesWrigley:
Alrighty, crashes are fixed by 2bef0ac and we're back in the green. The problem was using the old thread-unsafe way of stopping the heartbeat thread, which was causing assert failures in ZMQ.

JamesWrigley:
I would like to get this over the finish line, so my plan is to use the branch myself exclusively for a few weeks to see if there's any lingering bugs (@MasonProtter, @halleysfifthinc, if you'd like to try it too that'd also be amazeballs :heart:), and then merge it. @stevengj, @fredrikekre, would that be ok? I appreciate that it's a bit of a monster so if there's anything I can do to make reviewing easier lemme know and I'll try to help :sweat_smile:

JamesWrigley:
I decided to also try running JET over it, and a good thing I did because it uncovered quite a few bugs :grimacing: Some new, some old; fixes are in 8ff38d6.

JamesWrigley:
Gah, that's annoying. I was hoping we could get away without making this a breaking release but if there's a lot of packages depending on internals maybe it'd be safer. On the other hand in this particular case it wouldn't help anyway since IJulia isn't a dependency of ProgressMeter.jl.

JamesWrigley:
Sweet, thanks for the info :+1: I also haven't encountered more issues, apart from one with Bonito. My plan is:
1. Push the fixes for Bonito.
2. Merge this PR.
3. Get that ProgressMeter.jl PR merged.
4. Release a new IJulia version.
5. Make a Discourse post to inform people about the change.
6. Wait for the bug reports to roll in :bug: :trollface:

PR #1147, opened 2025-02-24
Fix latex symbol completion

yuyichao:
Broken by JuliaLang/julia#54800


yuyichao:
No `_completion_text(c::BslashCompletion) = c.bslash` was removed.

PR #1155, opened 2025-05-12
Do not append version info to manual `specname`

### LINKED ISSUE #1055: Warn or error for Invalid kernel names
I created a kernel with parantheses `(...)` in its name, and it seemed to work fine until I tried to start Jupyter with it. 
```
UserWarning: Invalid kernelspec directory name (Kernel names can only contain ASCII letters and numbers and these separators: - . _ (hyphen, period, and underscore).)
```
[Discourse thread (from someone else) with the same issue](https://discourse.julialang.org/t/weird-ijulia-issue-help-i-really/54877) 

Since Jupyter helpfully lists the characters allowed in the name, it would be better to alert the user with a warning, or better an error, instead of silently doing the wrong thing. The last message in #614 hints at the fact that you can use `specname` to circumvent this, but the user first has to know that this is an issue, and that this can be a solution. An error message seems a good way to convey this information, and to allow the user to quickly make the correction, rather than create the kernel and find out much later that it has this problem.

### LINKED ISSUE #1154: Julia version information is improperly appended to custom `specname`
When giving a custom `specname` as a keyword option, the Julia major and minor version (and a potential debug flag) are still appended. This should not happen: A manually given `specname` should be used verbatim (up to validity checks, cf #1055).

Otherwise, it is very hard to automate (in a `Makefile`, with `make install`, say) the creation of a custom project-specific kernel that should then later be uninstalled with `make uninstall`. With the current behavior, the `Makefile` would have to keep track of which Julia version was used during `make install` to then properly call `jupyter kernelspec uninstall <specname>`. Generally, it can't, because the Julia version might have changed between `make install` and `make uninstall`. Hence, it has to use a fixed, hardcoded `specname`.

Instead of unconditionally appending the version information to `specname = kernelspec_name(name)` as currently happens, the version information should be added only _inside_ `kernelspec_name`. That is, the version information is only included in the _default_ `specname` derived from `name`, but not in a custom given `specname`.

Note that the similar `display_name` already handles this correctly, and it would be good for `kernelspec_name` and `display_name` to behave consistently – in addition to the principle that when overriding the automatic `specname` by giving a keyword option, the computer should then not further override the user's choice.

goerz:
If `specname` is given manually, it should be used verbatim. By default, `specname` is still derived from `name` and the Julia version information. However, appending the version information now happens inside `kernelspec_name` (mirroring the implementation for `display_name`)

This also checks that any `specname` (manual or automatic) is restricted to containing ASCII letters, ASCII numbers, and the simple separators: `-` hyphen, `.` period, `_` underscore.

Closes #1154
Closes #1055


test/install.jl
```diff
@@ -5,6 +5,7 @@ import IJulia, JSON
     let kspec = IJulia.installkernel("ijuliatest", "-O3", "-p2",
                     env=Dict("FOO"=>"yes"), specname="Yef1rLr4kXKxq9rbEh3m")
         try
+            @test endswith(kspec, "Yef1rLr4kXKxq9rbEh3m")  # should not contain Julia version suffix
```
JamesWrigley, line 8: Should this be testing `==` equality since we want the final name to be strictly equal to `specname`? (modulo debug builds)

JamesWrigley (changes requested):

test/install.jl
goerz: If we split off the last part `kspec` (which I think is the full path), then yes. Sure, that sounds good!

JamesWrigley: Ah yes sorry, I guess it should be comparing `basename(kspec)`.

deps/kspec.jl
```diff
@@ -110,9 +113,13 @@ function installkernel(name::AbstractString, julia_options::AbstractString...;
     # Is IJulia being built from a debug build? If so, add "debug" to the description.
     debugdesc = ccall(:jl_is_debugbuild,Cint,())==1 ? "-debug" : ""
```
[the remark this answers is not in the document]
goerz: Yes! Thanks for catching that!

goerz:
All amended.

goerz:
Ugh, sorry for the noise. Screwed up `git add`. Let's try this again; should be correct, now.

JamesWrigley (approved):
LGTM!

Issue #1152, opened 2025-05-08
"--project=@." in argv in Julia kernelspecs is a bug


lewisl:
When this line is included in a kernelspec like so:

```
{
  "display_name": "Julia 1.11.5",
  "argv": [
    "/Users/myhomedirname/.julia/julia-1.11.5+0.aarch64.apple.darwin14/bin/julia",
    "-i",
   "--project=@.",
    "--color=yes",
    "/Users/myhomedirname/.julia/packages/IJulia/XF6bn/src/kernel.jl",
    "{connection_file}"
  ],
  "language": "julia",
  "env": {},
  "interrupt_mode": "signal"
}
```
jupyter will never be able to connect to the kernel and julia can't be run in that notebook.
That line means nothing to jupyter, it means something to Julia so this is Julia's bug.  I suspect the problem is that invoking the current directory as a project only works when that directory IS, in fact, a project.  But, that line won't create a project in the directory (e.g.--project.toml is not automagically created).  Only using Pkg or the package REPL via ] can do that.

I haven't tested this; it is a hunch.

In any case there is another issue in this repo that comments on the line being _newly put back_ in kernels sometime around Julia 1.11 or one of its maintenance releases (hundredths fraction in version number).  I suspect it had been removed earlier for a reason--it did not work.

Let's fix the bug and remove it.  People who want to run Julia jupyter notebooks as part of a project know how to use projects and environments, but might not do so for every notebook they use--so such folks also may want their notebooks to always connect.  One could do ```activate .``` as the first line in a notebook. Again, just a hunch.  In any case, no one loses any functionality by excluding the problem line from the argv vector. Activating a project environment just needs to be done in another way.  Or use Pluto if you never, ever need a Python notebook.

JamesWrigley:
This is very much by design: https://julialang.github.io/IJulia.jl/stable/manual/usage/#Julia-projects
By default IJulia is designed to be installed in the global environment and uses this mechanism to set the project semi-automatically, though of course one can create custom kernels that use a specific environment without having to have IJulia installed in the global environment. Have you been having issues with it?

JamesWrigley:
Relevant issue: #1127.

lewisl:
Except it doesn’t work.  And, previously the opposite approach had been adopted.

I found a pretty elaborate technical explanation why the failure to connect occurs. Briefly, Julia gets in a race condition trying to instantiate the project environment, which it can’t do, and Jupyter times out.

[...]

Issue #983, opened 2021-02-06
Don't install kernel by referencing specific kernel.jl file


twavv:
Seems like we could just change the kernelspec to have

```js
{
  "argv": [
    "/path/to/julia",
    "-i",
    "--startup-file=yes",  
    "--color=yes",
    "-e",
    "using IJulia; IJulia._launch_kernel()",
    "{connection_file}"
  ],
  ...
}
```

where we define a `_launch_kernel` command to just be the current contents of `kernel.jl`.

This would get around issues when using the wrong version of the kernel.jl file (e.g., when you've installed IJulia for a specific environment).

stevengj:
Yes, I think that should work.

Issue #950, opened 2020-09-22
IJulia not precompling within the process that jupyter launches


mauro3:
On Linux, launching jupyter form the shell with `jupyter notebook` and launching a new Julia 1.5 notebook from the browser prints the following in shell:
```
[ Info: Precompiling IJulia [7073ff75-c697-5162-941a-fcdaad2a7d2a]                                                      
ERROR: LoadError: ArgumentError: Package ZMQ does not have ZeroMQ_jll in its dependencies:                              
- If you have ZMQ checked out for development and have                                                                  
  added ZeroMQ_jll as a dependency but haven't updated your primary                                                     
  environment's manifest file, try `Pkg.resolve()`.                                                                     
- Otherwise you may need to report an issue with ZMQ                                                                    
Stacktrace:                                                                                                             
 [1] require(::Module, ::Symbol) at ./loading.jl:906                                                                    
 [2] include(::Function, ::Module, ::String) at ./Base.jl:380                                                           
 [3] include(::Module, ::String) at ./Base.jl:368                                                                       
 [4] top-level scope at none:2                                                                                          
 [5] eval at ./boot.jl:331 [inlined]                                                                                    
 [6] eval(::Expr) at ./client.jl:467                                                                                    
 [7] top-level scope at ./none:3                                                                                        
in expression starting at /home/mauro/.julia/packages/ZMQ/R3wSD/src/ZMQ.jl:4                                            
ERROR: LoadError: Failed to precompile ZMQ [c2297ded-f4af-51ae-bb23-16f91089e4e1] to /home/mauro/.julia/compiled/v1.5/ZM
Q/CaaO6_BNOax.ji.                                                                                                       
Stacktrace:                                                                                                             
 [1] error(::String) at ./error.jl:33                                                                                   
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1305                                                          
 [3] _require(::Base.PkgId) at ./loading.jl:1030                                                                        
 [4] require(::Base.PkgId) at ./loading.jl:928                                                                          
 [5] require(::Module, ::Symbol) at ./loading.jl:923                                                                    
 [6] include(::Function, ::Module, ::String) at ./Base.jl:380
 [7] include(::Module, ::String) at ./Base.jl:368
 [8] top-level scope at none:2
 [9] eval at ./boot.jl:331 [inlined]
 [10] eval(::Expr) at ./client.jl:467
 [11] top-level scope at ./none:3
[...]

PR #1156, opened 2025-05-12
Add kernel troubleshooting instructions

goerz:
This adds a section under "Troubleshooting" that explains how kernel specs work. I think a basic understanding of how this works goes a long way to identify all kinds of weird connection issues.

I feel like these kinds of instructions would probably help with issues like #1152 and #1152. Personally, whenever something goes wrong with a Jupyter kernel (Julia or otherwise), my first response is always to look at the spec.

@LilithHafner @lewisl Any suggestions for improvements?

### DIFF
--- a/docs/src/manual/troubleshooting.md
+++ b/docs/src/manual/troubleshooting.md
@@ -1,22 +1,5 @@
 # Troubleshooting
 
-
-## General troubleshooting tips
-
-* If you ran into a problem with the above steps, after fixing the
-  problem you can type `Pkg.build()` to try to rerun the install scripts.
-* If you tried it a while ago, try running `Pkg.update()` and try again:
-  this will fetch the latest versions of the Julia packages in case
-  the problem you saw was fixed.  Run `Pkg.build("IJulia")` if your Julia version may have changed.  If this doesn't work, you could try just deleting the whole `.julia/conda` directory in your home directory (on Windows, it is called `Users\USERNAME\.julia\conda` in your home directory) via `rm(abspath(first(DEPOT_PATH), "conda"),recursive=true)` in Julia and re-adding the packages. 
-* On MacOS, you currently need MacOS 10.7 or later; [MacOS 10.6 doesn't work](https://github.com/JuliaLang/julia/issues/4215) (unless you compile Julia yourself, from source code).
-* Internet Explorer 8 (the default in Windows 7) or 9 don't work with the notebook; use Firefox (6 or later) or Chrome (13 or later).  Internet Explorer 10 in Windows 8 works (albeit with a few rendering glitches), but Chrome or Firefox is better.
-* If the notebook opens up, but doesn't respond (the input label is `In[*]` indefinitely), try creating a new Python notebook (not Julia) from the `New` button in the Jupyter dashboard, to see if `1+1` works in Python.  If it is the same problem, then probably you have a [firewall running](https://github.com/ipython/ipython/issues/2499) on your machine (this is common on Windows) and you need to disable the firewall or at least to allow the IP address 127.0.0.1.  (For the [Sophos](https://en.wikipedia.org/wiki/Sophos) endpoint security software, go to "Configure Anti-Virus and HIPS", select "Authorization" and then "Websites", and add 127.0.0.1 to "Authorized websites"; finally, restart your computer.) If the Python test works, then IJulia may not be installed in the global or default environment and you may need to install a custom Julia kernel that uses your required `Project.toml` (see [Julia projects](@ref)).
-* Try running `jupyter --version` and make sure that it prints `3.0.0` or larger; earlier versions of IPython are no longer supported by IJulia.
-* You can try setting `ENV["JUPYTER"]=""; Pkg.build("IJulia")` to force IJulia to go back to its own Conda-based Jupyter version (if you previously tried a different `jupyter`).
-
-
-## Inspecting and fixing Jupyter kernels
-
 Jupyter notebooks run code by connecting to a [kernel](https://docs.jupyter.org/en/stable/projects/kernels.html). That's exactly what the `IJulia` package provides. Jupyter knows about available kernels through [kernel specs](https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs), which are `kernel.json` files inside [Jupyter's data paths](https://docs.jupyter.org/en/stable/use/jupyter-directories.html#data-files). You can list the available kernel specs on the command line with
 
 ```sh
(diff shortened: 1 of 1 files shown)


goerz:
Incidentally, https://github.com/JuliaLang/IJulia.jl/issues/983 seems like a good idea to me and would eliminate one category of potential errors, where the path to the `kernel.jl` file breaks

JamesWrigley:
Very nice work :heart: What do you think about moving those troubleshooting steps to the top of the page along with the rest of the tips? https://julialang.github.io/IJulia.jl/dev/manual/troubleshooting/#General-troubleshooting-tips

I feel that these things are the most likely to cause issues so it might be best to place them more prominently.

> Incidentally, #983 seems like a good idea to me and would eliminate one category of potential errors, where the path to the `kernel.jl` file breaks

Huh I missed that issue, yes we should absolutely do that :+1:

goerz:
Sure, I’ll have a look tomorrow. Do you mean just putting the new section first, to be followed by the (currently first) section “General troubleshooting tips”? Or somehow combine the two sections into a single section? The latter might be a bit weird; I’m not sure yet how I would title that combined section. But I’m sure I can come up with something. I have no particular strong opinions on how exactly any of this material should be organized.

JamesWrigley:
I was thinking of something like:
```
# Troubleshooting
<describe how kernels work>

## Connection failing tips
<tips you wrote>

## General troubleshooting tips
<other tips>
```

goerz:
That was easy enough to re-arrange, but I should probably wait for #1158 to be merged first, since that new approach will change the text significantly

docs/src/manual/troubleshooting.md
```diff
@@ -1,5 +1,49 @@
...
+
+
+## Kernel connection failure tips
+
+Fundamentally, if the `IJulia` kernel fails to connect, it might be due to any of the following issues:
+
+* The `julia` executable no longer exists (maybe you updated your installed Julia versions).
+* The environment that the `julia` executable runs in does not have the `IJulia` package installed. This is a common error. In general, the `IJulia` package should be installed in the base environment of your Julia installation (what you get when you type `] activate` into the REPL without any further options, or when you simply start the Julia REPL without any options). Note that the `--project=@.` option in the above `kernel.json` falls back to the base environment, so it should generally be safe. If you like to use [shared environments](https://pkgdocs.julialang.org/v1/environments/#Shared-environments), you may want to have a `--project` flag that references that shared environment, and make sure that `IJulia` is installed in that environment.
```
JamesWrigley, line 41: Could we some instructions for checking that the environment is fully instantiated? That seems to have tripped up a few people.

docs/src/manual/troubleshooting.md
```diff
@@ -1,5 +1,49 @@
...
+* The path to the `kernel.jl` file may no longer be valid (maybe due to installing an update of the `IJulia` package).
```
JamesWrigley, line 42: This can be removed now :partying_face:

docs/src/manual/troubleshooting.md
```diff
@@ -1,5 +1,49 @@
...
+  "interrupt_mode": "signal"
+}
+```
+
+Note the reference to the `julia` executable in line 4, and the reference to the `kernel.jl` file in line 8. There isn't much magical about kernels. All that happens when Jupyter starts the kernel based on a specific kernel spec is that it runs the process given by `argv`. That is, it runs `julia` with the given command line arguments. It then expects that it can talk to the resulting process with a specific [messaging protocol](https://jupyter-client.readthedocs.io/en/latest/messaging.html#messaging). Here, the code in `kernel.jl` exposes the implementation of that protocol.
+
+
+## Kernel connection failure tips
+
+Fundamentally, if the `IJulia` kernel fails to connect, it might be due to any of the following issues:
```
JamesWrigley, line 38: Could we add a tip about checking the notebook logs to see any error messages? For people running Jupyter from `jupyterlab()` or `notebook()` we can also mention `verbose=true`.

JamesWrigley (changes requested):

docs/src/manual/troubleshooting.md
goerz: I think there was some general confusion around the `--project=@.` flag, but it seems like a non-instantiated environment does not actually cause any problems with connecting to the kernel. I just tried this:

* Set up a kernel with `import IJulia; IJulia.installkernel("Julia", "--project=@.")`
* Create an empty directory for a new "notebook project"
* Initialize that empty directory by starting Julia with `julia --project=.`, and do `] add Plots` to create a `Project.toml` (and `Manifest.toml`) file
* Exit Julia and delete the `Manifest.toml` file. That should make the folder "not instantiated", right?
* Start Jupyter in that folder, by running `jupyter lab` from the command line (I have Jupyter installed independently from anything Julia-related in some Conda environment)
* Create a new notebook with the "Julia 1.11.5" kernel that we previously created

The kernel connects fine, also if I restart Jupyter and open the now existing notebook file. Of course, if I put `using Plots` in a cell in that notebook, I get the error message

```
ArgumentError: Package Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```

shown in the notebook. But that's expected, has a clear error message, and does not impact the connection to the kernel.

Maybe I'm overlooking something, and if someone can demonstrate how a non-instantiated environment would cause a problem, I'd be happy to add instructions to address that.

goerz: On the other hand, there's probably nothing wrong with adding a sentence "Also make sure that the environment is instantiated."

JamesWrigley: Right, but that's for when IJulia is installed in the global environment whereas this advice would be more relevant for when it's installed in a different environment (which would be easier to miss being un-instantiated than the global environment). #950 is one example, but I think it's also happened to me in the past with stale manifests.

docs/src/manual/troubleshooting.md
goerz: Yes, I've rewritten this to:

> Fundamentally, if the `IJulia` kernel fails to connect, it is most likely due to one of the following two issues:
>
> * The `julia` executable no longer exists (maybe you updated your installed Julia versions).
> * The environment that the `julia` executable runs in does not have the `IJulia` package installed. This is a common error. In general, the `IJulia` package should be installed in the base environment of your Julia installation (what you get when you type `] activate` into the REPL without any further options, or when you simply start the Julia REPL without any options). Note that the `--project=@.` option in the above `kernel.json` falls back to the base environment, so it should generally be safe. If you like to use [shared environments](https://pkgdocs.julialang.org/v1/environments/#Shared-environments), you may want to have a `--project` flag that references that shared environment, and make sure that `IJulia` is installed in that environment. Also make sure that the environment is instantiated.

goerz:
Pushed a new commit that should address all the requested changes, as discussed above

JamesWrigley (approved):
LGTM!

PR #1158, opened 2025-05-13
Move kernel.jl into a function in the IJulia module

### LINKED ISSUE #983: Don't install kernel by referencing specific kernel.jl file

JamesWrigley:
This means we don't need to hardcode a path to `kernel.jl` anymore. Also removed some old code from kernel.jl:
- The Apple `PATH` problem was fixed upstream in 2013.
- We set `-i` now so the call to `jl_exit_on_sigint()` and explicitly setting `Base.is_interactive` is no longer necessary (see the docstring for `Base.exit_on_sigint`).
- No need to delete the `SOURCE_PATH` TLV since we can `include()` into a module now.

Fixes #983. Tested manually.

(CC @goerz)

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -1,53 +1,32 @@
-import IJulia
+function run_kernel()
+    # Load InteractiveUtils from the IJulia namespace since @stdlib might not be in LOAD_PATH
+    @eval Main using IJulia.InteractiveUtils
 
-let
-    ijulia_kernel_file = joinpath(dirname(pathof(IJulia)), "kernel.jl")
-    this_kernel_file = @__FILE__
-    if stat(ijulia_kernel_file) != stat(this_kernel_file)
-        @warn "this kernel.jl is different from the one provided by IJulia" this_kernel_file ijulia_kernel_file
+    # the size of truncated output to show should not depend on the terminal
+    # where the kernel is launched, since the display is elsewhere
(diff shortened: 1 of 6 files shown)


src/kernel.jl
```diff
@@ -1,53 +1,32 @@
...
-IJulia.watch_stdio()
+    # import things that we want visible in IJulia but not in REPL's using IJulia
+    @eval Main import IJulia: ans, In, Out, clear_history
 
-# workaround JuliaLang/julia#4259
-delete!(task_local_storage(),:SOURCE_PATH)
+    pushdisplay(IJulia.InlineDisplay())
 
-# workaround JuliaLang/julia#6765
-Core.eval(Base, :(is_interactive = true))
```
stevengj, line 44: What happened to `is_interactive = true`?

oschulz:
Merci @JamesWrigley !

Issue #1099, opened 2023-11-30
Autocomplete problem in output cell.


garazha-ilya:
Autocompletion works well, but errors inside output cell occured every time. 

1. The output of `versioninfo()` :

Julia Version 1.9.3
Commit bed2cd540a1 (2023-08-24 14:43 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-11370H @ 3.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, tigerlake)
  Threads: 1 on 8 virtual cores

2. How you installed Julia

```
wget https://julialang-s3.julialang.org/bin/linux/x64/1.9/julia-1.9.4-linux-x86_64.tar.gz
tar zxvf julia-1.9.4-linux-x86_64.tar.gz
```
3. A minimal working example (MWE), also known as a minimum reproducible example:
 `using Pkg; Pkg.add("IJulia"); using IJulia; notebook()`

Settings -> Settings Editor -> Code Completion -> Enable autocompletion -> true

In first evaluation (Shift + Enter) everything works fine, but after second evaluation of the same code occurs error message:

![image](https://github.com/JuliaLang/IJulia.jl/assets/151720612/7357fa15-b411-45f1-bb46-9dd9ed65ca7f)

If autocompletion is disabled, then everything works normal.

```
KERNEL EXCEPTION
BoundsError: attempt to access empty String at index [1:40]

Stacktrace:
 [1] checkbounds
   @ ./strings/basic.jl:216 [inlined]
 [2] getindex(s::String, r::UnitRange{Int64})
   @ Base ./strings/string.jl:276
 [3] complete_request(socket::ZMQ.Socket, msg::Msg)
   @ IJulia ~/.julia/packages/IJulia/Vo51o/src/handlers.jl:115
 [4] #invokelatest#2
   @ ./essentials.jl:819 [inlined]
 [5] invokelatest
   @ ./essentials.jl:816 [inlined]
 [6] eventloop(socket::ZMQ.Socket)
   @ IJulia ~/.julia/packages/IJulia/Vo51o/src/eventloop.jl:8
 [7] (::IJulia.var"#15#18")()
   @ IJulia ./task.jl:514
```

stevengj:
[...]

PR #1159, opened 2025-05-16
Make `get_token(code, pos)` handle `pos` being out of bounds

JamesWrigley:
This would previously cause inspection requests to fail when autocompletion was enabled.

The equivalent function in IPython seems to end up doing this implicitly: https://github.com/ipython/ipython/blob/5a3bf345ed7c78e5fab84faa81285e5fd9dfa3da/IPython/utils/tokenutil.py#L196


JamesWrigley:
Did the same fix for `completion_request()` in e35c9a7, which should hopefully fix the exception reported in #1099.

JamesWrigley:
(if there's no objections I'll merge this at the end of the week)

PR #1160, opened 2025-05-17
Strip whitespace from Pkg commands

### LINKED ISSUE #1095: Support parsing of comma separated packages in `] add`
This has been supported in `Pkg` REPL mode in the actual REPL for a while but fails in IJulia:

image

JamesWrigley:
This avoids a parsing error in Pkg with `] add Foo, Bar`.

Fixes #1095.


stevengj (approved):

stevengj:
Sounds good, though arguably Pkg parsing shouldn't be so brittle.

JamesWrigley:
Yeah, I opened an issue for it here: https://github.com/JuliaLang/Pkg.jl/issues/4239

PR #1163, opened 2025-06-02
Fix a deadlock in `shutdown_requests`

JamesWrigley:
I noticed when testing locally that the kernel would consistently get SIGTERM'd when restarted, turns out that it was always getting into a deadlock when trying to send the `shutdown_reply` message and thus the notebook was forcefully killing it after a timeout. @halleysfifthinc, I wonder if this is the error you were seeing in #1135? The stacktrace from the heartbeat thread is printed too and it's easy to miss the stacktrace from the main thread.

This is a fairly serious bug so if there's no objections I'll merge and release it later today.

### DIFF
--- a/src/handlers.jl
+++ b/src/handlers.jl
@@ -229,6 +229,14 @@ function shutdown_request(socket, msg)
     # stop heartbeat thread
     stop_heartbeat(heartbeat[], heartbeat_context[])
 
+    # Shutdown the `requests` socket handler before sending any messages. This
+    # is necessary because otherwise the event loop will be calling
+    # `recv_ipython()` and holding a lock on `requests`, which will cause a
+    # deadlock when we try to send a message to it from the `control` socket
+    # handler.
+    global _shutting_down[] = true
+    @async Base.throwto(requests_task[], InterruptException())
+
     send_ipython(requests[], msg_reply(msg, "shutdown_reply",
                                        msg.content))
     sleep(0.1) # short delay (like in ipykernel), to hopefully ensure shutdown_reply is sent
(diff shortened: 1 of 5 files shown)


src/handlers.jl
```diff
@@ -229,6 +229,14 @@ function shutdown_request(socket, msg)
     # stop heartbeat thread
     stop_heartbeat(heartbeat[], heartbeat_context[])
 
+    # Shutdown the `requests` socket handler before sending any messages. This
+    # is necessary because otherwise the event loop will be calling
+    # `recv_ipython()` and holding a lock on `requests`, which will cause a
+    # deadlock when we try to send a message to it from the `control` socket
+    # handler.
+    global _shutting_down[] = true
+    @async Base.throwto(requests_task[], InterruptException())
```
halleysfifthinc, line 238: I think a contributing factor to this bug is that the "shutdown_reply" being sent on the SHELL/requests channel instead of the CONTROL channel. My reading of the client messaging spec is that all replies should be sent on the channel they were received on, so I believe that

```suggestion
    @async Base.throwto(socket, InterruptException())
```

or 
```suggestion
    @async Base.throwto(control[], InterruptException())
```
would be more correct (I've never seen a `shutdown_request` sent on the SHELL channel, and that using the SHELL channel for `shutdown_request`s is deprecated on 5.4 of the client messaging API) and likely avoid a deadlock from the SHELL/requests channel being locked.

halleysfifthinc (commented):
I'm not sure, but I think this is probably a different bug than what I had noticed.

PR #1177, opened 2025-08-25
Load Pkg.jl lazily to cut down on load times

JamesWrigley:
Also bump the version.


stevengj:
LGTM, but contains a lot of changes besides the change to loading of Pkg?

JamesWrigley:
Figured it out, the `minirepl` global variable was holding a reference to `stdout`, which was an `IJuliaStdio` object holding a reference to a `Kernel`, which would get saved in the pkgimage and cause crashes. Fixed in 6fc832e by making it part of the `Kernel` struct instead of a global.

PR #1178, opened 2025-08-27
Adding default value for `set_max_stdio` keyword argument `kernel`

quffaro:
`set_max_stdio` seems to be one of the few (or only) functions that doesn't have a default value for `kernel`, so I added the `_default_kernel` as its default value. I've verified that tests pass.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "IJulia"
 uuid = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
-version = "1.30.0"
+version = "1.30.1"
 
 [deps]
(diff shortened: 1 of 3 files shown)


JamesWrigley (changes requested):
Whoops thanks, this was accidentally a breaking change. Could you also bump the version and add a changelog entry?

quffaro:
I've added a changelog entry, but since there was no issue connected to this PR I didn't include a link.

JamesWrigley:
Ah actually this should link to the PR, you can add `([#1178])` to the changelog sentence and Changelog.jl will insert a link.

PR #1186, opened 2025-10-03
Add support for JSON 1.0 release

quinnj:


JamesWrigley:
Not sure about this since JSON v1 is a heavy-ish dependency. I was wondering if we should vendor JSONX instead, and then if we load Conda lazily we can avoid both Conda and JSON. Are you aware of any performance limitations with JSONX? @stevengj, what do you think?

JamesWrigley:
One downside is that JSONX parses all numbers as floats, so we would need to be careful to convert the appropriate fields to ints.

quinnj:
v1 isn't going to be much/any heavier than pre-1.0; it's certainly still lighter than ZMQ. The StructUtils/PrecompileTools deps are pretty minimal (and IJulia already deps on PT). That said, JSONX perf is pretty decent for being very simple.

You're going to get a much bigger gain from doing Conda lazily than JSON.

JamesWrigley:
After some consideration, I'm inclined to vendor JSONX. I'll have a shot at that and do some benchmarks. @stevengj, @fredrikekre, any thoughts?

PR #1197, opened 2025-10-17
Support JSON v1

hhaensel:
define a local function `parsejson()` that parses into "Dict{String,Any}" for both JSON < v1 and JSON v1

### DIFF
--- a/docs/src/_changelog.md
+++ b/docs/src/_changelog.md
@@ -9,6 +9,11 @@ Changelog](https://keepachangelog.com).
 
 ## Unreleased
 
+## [v1.31.1] - 2025-10-20
+
+### Added
+- Added support for JSON v1 ([#1197]).
+
 ### Fixed
 - Fixed support for 32bit systems ([#1196]).
 
--- a/src/IJulia.jl
+++ b/src/IJulia.jl
@@ -34,7 +34,7 @@ module IJulia
 export notebook, jupyterlab, installkernel
 
 import SHA
-using ZMQ, JSON
+using ZMQ
 import Base: invokelatest, RefValue
 import Dates
 using Dates: now, format, UTC, ISODateTimeFormat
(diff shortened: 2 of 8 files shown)


src/IJulia.jl
```diff
@@ -47,6 +47,12 @@ import Logging
 # InteractiveUtils is not used inside IJulia, but loaded in src/kernel.jl
 # and this import makes it possible to load InteractiveUtils from the IJulia namespace
 import InteractiveUtils
+import JSON.json
+@static if isdefined(JSON, :Object)
```
JamesWrigley: I have a very mild preference for using `if pkgversion(...) < v"..."`, but this is also ok if we include a comment.

hhaensel:
I must admit that I'm not familiar with Changelog and how to add it., I had searched the Web a bit but I'm not completely sure how to do it. Could you tell me how that works in your case? Happy to learn this.

JamesWrigley:
All good :) You just need to add an entry in this file: https://github.com/JuliaLang/IJulia.jl/blob/master/docs/src/_changelog.md
You could also go ahead and bump the version to v1.31.1 to be ready for when Conda.jl is updated (it's a bit weird that our compat already includes JSON 1 but isn't actually compatible with it...).

stevengj:
Couldn't we just support the `AbstractDict` type returned by `JSON.parse` instead of converting to a `Dict`?

docs/src/_changelog.md
```diff
@@ -9,9 +9,14 @@ Changelog](https://keepachangelog.com).
 
 ## Unreleased
 
+## [v1.31.1] - 2025-10-18
+
 ### Fixed
 - Fixed support for 32bit systems ([#1196]).
 
+### Added
+- Added support for JSON v1
```
JamesWrigley: This should also link to the PR:
```suggestion
- Added support for JSON v1 ([#1197]).
```

src/IJulia.jl
```diff
@@ -47,6 +47,12 @@ import Logging
 # InteractiveUtils is not used inside IJulia, but loaded in src/kernel.jl
 # and this import makes it possible to load InteractiveUtils from the IJulia namespace
 import InteractiveUtils
+import JSON.json
```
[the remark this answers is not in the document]
hhaensel: Yes, we use it in line 50 - 53 of msg.jl

hhaensel:
Had forgotten to remove the json import, therefore forcepushed the branch once more. Should be ok now, sorry.

PR #1209, opened 2025-11-02
permit more python versions

Krastanov:
Many important scientific packages do not yet work on python 3.13. By relaxing this constraint, wrappers like PyQDecoders.jl can be used together with IJulia.

Tested locally.

Before this change `add PyQDecoders IJulia` would fail at the conda resolve step.


JamesWrigley:
Hmm the PythonCall extension doesn't actually depend on a particular Python version, the `CondaPkg.toml` file is just there for the tests. Maybe we can put it somewhere else so that CondaPkg doesn't use it at all?

PR #1211, opened 2025-11-03
make an ijulia "app" entrypoint

### LINKED ISSUE #1236: Should `IJulia` be an app?
So that one can do `/> ijulia -notebook` or whatever directly rather than `/> julia` followed by `using IJulia; notebook()`?

KristofferC:
this makes ijulia installable via the Pkg app interface: https://pkgdocs.julialang.org/v1/apps/ and allows one to do e.g.

```
pkg> app add IJulia
```

and then run e.g.

```
ijulia lab
```

in the terminal to start an ijulia jupyterlab notebook without having to launch julia itself and load IJulia explicitly.

To test this PR you can do `Pkg.Apps.develop(path=abspath("."))` (the relative path is currently bugged), make sure `.julia/bin` is in the PATH and then run `ijulia`.

### DIFF
--- a/docs/src/manual/running.md
+++ b/docs/src/manual/running.md
@@ -3,11 +3,13 @@
 
 ## Command-Line Launcher
 
-Starting with Julia 1.12, you can launch IJulia directly from the command line using the `ijulia` command. First, you need to install the app entry point by running in the Julia REPL:
+Starting with Julia 1.12, you can launch IJulia directly from the command line using the `ijulia` command. First, you need to install the app entry point (see [Julia app documentation](https://pkgdocs.julialang.org/v1/apps/)) by running in the Julia REPL:
 
 ```julia
 pkg> app add IJulia
 ```
+
+You may need to add `~/.julia/bin` to your PATH if it's not already there.
 
 Then you can use the `ijulia` command from your terminal:
 
--- a/docs/src/manual/running.md
+++ b/docs/src/manual/running.md
@@ -1,6 +1,45 @@
 # Running IJulia
 
 
+## Command-Line Launcher
+
+Starting with Julia 1.12, you can launch IJulia directly from the command line using the `ijulia` command. First, you need to install the app entry point by running in the Julia REPL:
+
+```julia
+pkg> app add IJulia
+```
+
+Then you can use the `ijulia` command from your terminal:
+
+```bash
+ijulia
+```
+
+This provides a convenient way to launch Jupyter without starting a Julia REPL session first. The launcher supports the following options:
+
+- `ijulia` or `ijulia notebook` - Launch Jupyter Notebook (default)
+- `ijulia lab` - Launch JupyterLab
+- `--dir=PATH` - Launch in the specified directory (default: home directory)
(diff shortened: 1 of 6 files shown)


src/IJulia.jl
```diff
@@ -584,4 +572,92 @@ include("inline.jl")
 include("kernel.jl")
 include("precompile.jl")
 
+#######################################################################
+# App entry point for `ijulia` command
+
+# Mutable function references for testing - allows tests to mock these functions
+notebook_cmd::Function = notebook
+jupyterlab_cmd::Function = jupyterlab
```
JamesWrigley: Is it possible to pass these in as arguments to `@main`? I slightly prefer that over global variables.

test/main.jl
```diff
@@ -0,0 +1,127 @@
...
+        ret = @test_logs (:error, r"Unknown subcommand.*Use 'notebook' or 'lab'") IJulia.main(["invalid"])
+        @test ret == 1
+    end
+
+    # Test argument parsing without actually launching
+    # We use mutable function references (notebook_cmd/jupyterlab_cmd) to test parsing
+    @testset "argument parsing" begin
+        # Save original functions
+        orig_notebook = IJulia.notebook_cmd
+        orig_jupyterlab = IJulia.jupyterlab_cmd
```
JamesWrigley: Nitpicking, I'd move these to the toplevel testset so each sub-testset doesn't need to store them.

docs/src/manual/running.md
```diff
@@ -1,6 +1,45 @@
 # Running IJulia
 
 
+## Command-Line Launcher
+
+Starting with Julia 1.12, you can launch IJulia directly from the command line using the `ijulia` command. First, you need to install the app entry point by running in the Julia REPL:
+
+```julia
+pkg> app add IJulia
```
JamesWrigley: I think IJulia is one of the few packages that still has a build step 😅 Shall we just install the app automatically when building? My only concern is that from reading the docs it seems that an app is somehow hardcoded to a specific Julia binary? That's a bit unfortunate because we recently added a juliaup integration that reduced the need for rebuilding IJulia to once per minor release (#1201). And perhaps if it's that experimental we should leave it out by default, can always enable it later.

JamesWrigley (changes requested):
Looks like this will need to be gated behind a `@static if VERSION >= v"1.11"` to support 1.10.

docs/src/manual/running.md
JamesWrigley: That sounds ok, as long as it still installs the same kernels as now. @stevengj, what do you think?

KristofferC:
> How is build support broken? 👀

`app add` just doesn't run `build`... :laughing:

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/4491 should fix the app build issue.

But really, this whole build thing is kind of user hostile :laughing:. Like, there is code in here to just do what the user want but instead they get an error and have to do some `Pkg.build` etc. 

Anyway, I'll work on my no-build branch.

PR #1215, opened 2025-11-05
change from using a build script to auto installing kernels and providing a jupyter path update function

KristofferC:
The Pkg build functionality was made for a different era in Julia package management. It gets run whenever a package is *downloaded* (which is a weird thing to trigger on where packages are considered immutable and multiple environments share the same package downloads), it requires the package to recompile when build options are changed etc. There are now better options based on e.g. scratch spaces.

Instead of using `Pkg.build` to install the default kernel, this instead auto installs it if needed.
It provides a separate function to update the jupyter path because tying kernel installation and updating of jupyter path together felt wrong.

It moves to using the Preference system over using `DEPOT_PATH[1]/prefs` (but reads the old file as a fallback)

It changes some things that previously had to be done at build time (like setting the debug mode) to now just be a runtime thing.

Using IJulia from scratch (by using an empty depot and clearing out all kernels) looks like:

```
JULIA_DEPOT_PATH="$(mktemp -d):" julia +1.12 --project -q

(IJulia) pkg> instantiate 
...
  Installing artifacts ━━━━━━━━━━ 2/2
    Building Conda → `/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/tmp.8kRd4JRUqk/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/8f06b0cfa4c514c7b9546756dbae91fcfbc92dc9/build.log`

julia> using IJulia
Precompiling IJulia finished.
  13 dependencies successfully precompiled in 11 seconds. 30 already precompiled.

julia> notebook()
┌ Info: No default Julia kernel found for Julia 1.12.
│ Installing kernel automatically. You can reinstall or update the kernel
│ anytime by running: IJulia.installkernel()
│ 
│ To disable this auto-installation, set the environment variable:
└     ENV["IJULIA_NODEFAULTKERNEL"] = "true"
[ Info: Installing 'Julia 1.12' kernelspec in /Users/kc/Library/Jupyter/kernels/julia-1.12
...
install Jupyter via Conda, y/n? [y]: y
[ Info: Downloading miniconda installer ...
[ Info: Installing miniconda ...
PREFIX=/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/tmp.8kRd4JRUqk/conda/3/aarch64
Unpacking bootstrapper...
Unpacking payload...
...
...
[ Info: running setenv(`/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/tmp.8kRd4JRUqk/conda/3/aarch64/bin/jupyter notebook`)
...
```

### DIFF
--- a/src/jupyter.jl
+++ b/src/jupyter.jl
@@ -133,7 +133,7 @@ function run_subcommand(name, package_name, port, args...)
 end
 
 """
-    notebook(args=``; dir=homedir(), detached=false, port::Union{Nothing,Int}=nothing, verbose=false)
+    notebook(args=``; dir=homedir(), detached=false, port::Union{Nothing,Int}=nothing, verbose=ijulia_debug())
 
 The `notebook()` function launches the Jupyter notebook, and is
 equivalent to running `jupyter notebook` at the operating-system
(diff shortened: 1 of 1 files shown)


JamesWrigley (changes requested):
Thanks for making it so quickly ❤️ 

Also CC @goerz and @stevengj, in case you have any suggestions.

docs/src/_changelog.md
```diff
@@ -29,6 +35,19 @@ Changelog](https://keepachangelog.com).
   fixed ([#1202]).
 
 ### Changed
+- Removed `Pkg.build("IJulia")` support. IJulia no longer uses a
+  build step for kernel installation. The build-time configuration system has
+  been replaced with runtime functions. Use [`installkernel()`](@ref) to install
+  or update kernels, and [`update_jupyter_path()`](@ref) to configure the
+  Jupyter executable path.
+- IJulia now uses [Scratch.jl](https://github.com/JuliaPackaging/Scratch.jl) to
+  store configuration preferences instead of writing to `DEPOT_PATH/prefs/`.
```
stevengj: Why not Preferences.jl?

KristofferC:
Just FYI, while doing this I was not aware of the  juliaup issue being referenced (and this PR was not meant to address that) but from the comment I can sort of imagine what the problem is. Pkg Apps have the same problem in that we store the absolute path to the Julia executable that installed the path. When juliaup does GC that exe, it's sad times.

My plan to help with the Pkg App problem is by using https://github.com/JuliaLang/juliaup/pull/1315 in combination with a Pkg App specific env variable to override the executable that launch the app to be just `julia`. The automatic julia version selection will then pick the correct julia version by looking at the manifest, if needed download the needed julia version and launch the app

docs/src/_changelog.md
JamesWrigley: Wouldn't that make changing these settings force re-precompilation? I don't think we need that.

docs/src/_changelog.md
```diff
@@ -29,6 +35,19 @@ Changelog](https://keepachangelog.com).
   fixed ([#1202]).
 
 ### Changed
+- Removed `Pkg.build("IJulia")` support. IJulia no longer uses a
+  build step for kernel installation. The build-time configuration system has
```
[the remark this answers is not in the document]
KristofferC: We could but if `add IJulia` runs the build script or not is subject to the state of the file system. If the IJulia version is found in some existing DEPOT it won't build. So it feels kind of brittle to rely on this.

KristofferC:
Changed to using Preferences and did some updates based on review and own reading of code. I think the `update_jupyter_path` is quite important to look at now.

README.md
```diff
@@ -31,29 +31,20 @@ Install IJulia from the Julia REPL by pressing `]` to enter pkg mode and enterin
...
-for whether it should install Jupyter.  Hit enter to
-have it use the [Conda.jl](https://github.com/Luthaf/Conda.jl)
-package to install a minimal Python+Jupyter distribution (via
-[Miniconda](https://www.anaconda.com/docs/getting-started/miniconda/install#quickstart-install-instructions)) that is
-private to Julia (not in your `PATH`).
+The first time you run `notebook()`, it will:
+- Prompt you to install Jupyter if you don't already have it (hit enter to install via [Conda.jl](https://github.com/Luthaf/Conda.jl), which creates a minimal Python+Jupyter distribution private to Julia)
+- Automatically install the Julia kernel for your current Julia version
+
+If you already have Jupyter installed and prefer to use it, you can launch it from the terminal with `jupyter notebook` instead. The Julia kernel will be automatically installed the first time you run `IJulia.notebook()` or `IJulia.jupyterlab()`.
```
stevengj: It would be better to just tell people to run `installkernel()` if they want to use an externally installed Jupyter front-end of some kind.

JamesWrigley:
(bump)

README.md
JamesWrigley: Fixed in ab9137c.

PR #1220, opened 2025-11-07
TTFX improvements

JamesWrigley:
- Despecialize display methods
- Move Revise support into a package extension
- Add auto-precompilation after certain Jupyter requests

### DIFF
--- a/test/kernel.jl
+++ b/test/kernel.jl
@@ -492,8 +492,10 @@ end
 
                     # Note that we don't wait for a reply because the kernel
                     # will shut down almost immediately and it's not guaranteed
-                    # we'll receive the reply.
+                    # we'll receive the reply. We also sleep for a bit to try to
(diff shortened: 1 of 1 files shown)


ext/IJuliaReviseExt.jl
```diff
@@ -0,0 +1,35 @@
...
+end
+
+function __init__()
+    if @ccall(jl_generating_output()::Cint) == 0 && get(ENV, "JULIA_REVISE", "auto") == "auto"
+        IJulia.push_preexecute_hook(revise_hook)
+    end
+end
+
+@compile_workload begin
+    revise_hook()
```
KristofferC: Feel to me like an explicit `precompile` directive would work just as well as running the code in this case.

And does calling revise below really reduce TTFR (time to first revise).

PR #1226, opened 2025-11-14
remove Jupyter config from ARGS

### LINKED ISSUE #1225: `ARGS` is non-empty at the beginning of a session
I have some Julia files that can be used interactively or with command line arguments. Inside an IJulia notebook session, however, I cannot use them because they get confused by the non-empty `ARGS` vector:
```
In [1]: ARGS

Out[1]: 1-element Vector{String}:
 "/home/matthias314/.local/share/jupyt" ⋯ 28 bytes ⋯ "4a3-4b76-b107-8be22b6b3e81.json"
```
What about running `empty!(ARGS)` at the beginning of an IJulia session?

I'm using IJulia 1.32.1.

stevengj:
Closes #1225.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -10,6 +10,7 @@ function run_kernel()
     println(Core.stdout, "Starting kernel event loops.")
     kernel = Kernel()
     IJulia.init(ARGS, kernel)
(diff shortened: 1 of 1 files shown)


JamesWrigley (approved):
LGTM!

